PR #1, opened 2024-10-16
A naive attempt at a tensor type

seanmcl:
Mainly interested in feedback here. Tensors have some constraints, e.g. that the shapes are recursively consistent. I wanted to avoid using the shape in the type, so tried to bake well-formedness into the shape method. Not sure this is a good approach. Another thing I thought of was to write an "OK" method, and then have "OK" as an antecedent to all the definitions that require it.  Anyway, this is one naive way to define a tensor and I just wanted to get something off the ground to start implementing something like a fragment of numpy.


TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
 
 def hello := "world"
 
+class TensorElement (a: Type) where
+  add: a -> a -> a
```
jtristan, line 6: You should use those: https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Add.add
That way, you'll get the nice syntax and the theorems.

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
 
 def hello := "world"
```
jtristan, line 4: You should use a namespace

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
...
 
+class TensorElement (a: Type) where
```
jtristan, line 5: You may consider use the variable mechanism for a type like a, and also using {} to have it inferred. However, in this case, it may be good to start like this.

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
```
jtristan, line 2: I like to reserve the Basic.lean file to just import all the other relevant files in the directory. You may want to create two files, one for the tensor elements, and one for the tensor themselve.

TensorLib/Basic.lean
seanmcl: Ah ok, so we will have type classes for each primitive operation and then mix them together as needed for each more complex operation. Seems nice.

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
 
 def hello := "world"
 
+class TensorElement (a: Type) where
+  add: a -> a -> a
+  -- etc. enough to implement numpy ops
+
+inductive TensorData (a: Type) [TensorElement a] where
```
jtristan, line 9: Again, consider sections and variables

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
...
+| td_root (x: List a)
+| td_node (tl: List (TensorData a))
+
+-- Awkward... I'm sure this exists somewhere in the stdlib
+def sequence {a: Type} (x: List (Option a)): Option (List a) :=
```
jtristan, line 14: You need to familiarize yourself with monads in Lean, this kind of code will be less Awkward

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
...
+  | TensorData.td_node xs =>
+    match sequence (.map Shape xs) with
+    | none => none
+    | some .nil => some .nil
+    | some (.cons x xs) =>
+      if List.all xs (fun x' => x == x') then
+        some (List.cons (1 + List.length xs) x)
+      else
+        none
+
```
jtristan, line 36: We may want to pack a TensorData and a proof that Shape ... = some ... in a dependent record.

TensorLib/Basic.lean
```diff
@@ -2,4 +2,98 @@ namespace TensorLib
...
+  match x with
+  | .nil => some .nil
+  | .cons none _ => none
+  | .cons (some y) ys =>
+    match sequence ys with
+    | none => none
+    | some ys' => some (.cons y ys')
+
+-- somehow prove termination here
+def Shape {a: Type} [TensorElement a] (x: TensorData a): Option (List Nat) :=
```
jtristan, line 24: Question: partial function or relation? Maybe we should implement both to get a feel for the two options.

TensorLib/Basic.lean
seanmcl: I see what you mean. I'll be explicit/pedantic for now.

seanmcl:
Have a new cut now. I think I've addressed the comments.

TensorLib/Basic.lean
seanmcl: We'll work this out in the next PR.

TensorLib/Basic.lean
seanmcl: Ah didn't do this yet. Will add.

seanmcl: Added.

TensorLib.lean
```diff
@@ -1,3 +1,5 @@
 -- This module serves as the root of the `TensorLib` library.
 -- Import modules here that should be built as part of the library.
-import TensorLib.Basic
\ No newline at end of file
+import TensorLib.Basic
+import TensorLib.TensorElement
```
jtristan, line 4: This file should only import `TensorLib.Basic` which in turn will import `TensorLib.TensorElement` and `TensorLib.Tensor`

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
+import Mathlib.Tactic
```
jtristan, line 1: You're supposed to use a copyright notice such as: 
/-
Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
Released under Apache 2.0 license as described in the file LICENSE.
Authors: Foo Bar
-/

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
...
+
+-- I tried `td_node (x: TensorData) (xs: List TensorData)` to remove the redundant empty node
+-- from the type. (We can already represent the 0 shape with `td_root []`.)
+-- This design works, but this way the examples are more uniform; it looked weird to
+-- syntactically preference the first of every node list. The code has some annoying special
+-- cases for the empty node now, though, so it may be worth revisiting this.
+inductive TensorData where
+| td_root (xs: List a)
+| td_node (xs: List TensorData)
+  abbrev Shape := List Nat
```
jtristan, line 15: Indent

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
...
+
+-- TODO: Make tail-recursive, perhaps when Lean will eliminite tail calls in mutual-recursion
+def hasShape (x: TensorData a) (shape: Shape): Bool :=
```
jtristan, line 18: `def TensorData. hasShape ` ... to enable dot notation.

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
...
+    | List.cons n .nil => List.length y == n
+    | _ => false
+  | .td_node ys => match shape with
+    | List.cons n shape' =>
+      List.length ys == n &&
+      List.all ys (fun y => hasShape y shape')
+    | _ => false
+
+  -- Matching multiple patterns does not work well with the termination checker
+  -- I've asked about this on #lean-lang
```
jtristan, line 30: zulip

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
...
+      .td_root ([1, 2, 3]: List Nat),
+      .td_root ([4, 5, 6]: List Nat),
+      .td_root ([7, 8, 9]: List Nat),
+    ],
+    .td_root ([1, 2, 3]: List Nat),
+  ]
+)
+-- TODO
+theorem flattens_same : ∀ x, flatten x = flatten_mutual x := by
+  sorry
```
jtristan, line 132: We cannot merge a sorry into main

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
+import Mathlib.Tactic
+namespace TensorLib
```
jtristan, line 2: imports are following by the docstring

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
...
+section TensorData
+
+variable (a: Type)
+
+-- I tried `td_node (x: TensorData) (xs: List TensorData)` to remove the redundant empty node
+-- from the type. (We can already represent the 0 shape with `td_root []`.)
+-- This design works, but this way the examples are more uniform; it looked weird to
+-- syntactically preference the first of every node list. The code has some annoying special
+-- cases for the empty node now, though, so it may be worth revisiting this.
+inductive TensorData where
```
jtristan, line 12: Should we call this `Tensor`? That's how we refer to this datatype, and that's the name of the file.

TensorLib/Tensor.lean
jtristan, line 18: It may be best to define hasShape as a predicate, could be easier to reason about. Then we can still provide an implementation and how that they agree.

seanmcl: Neat.

TensorLib/Tensor.lean
seanmcl: Was thinking that the tensor would have some metadata in it, e.g. the shape.

TensorLib/Tensor.lean
```diff
@@ -0,0 +1,135 @@
+import Mathlib.Tactic
+namespace TensorLib
+section TensorData
+
+variable (a: Type)
```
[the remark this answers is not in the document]
seanmcl: I had it like this originally, but we don't use it anywhere. Do you think there's value to it? For example, if we get the `TensorElement` restrictions on the type for free then it may be worth it.

TensorLib/Tensor.lean
seanmcl: I did this below as an inductive predicate `HasShape`. Was that not what you meant?

jtristan: I missed it.

TensorLib/Tensor.lean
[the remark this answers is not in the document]
seanmcl: I was planning on adding the restriction anywhere I actually need it. Maybe that's essentially everwhere at some point so I should just restrict right off the bat.

TensorLib/Tensor.lean
seanmcl: Done

seanmcl: Done

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+        intro H2 H3
+        rw[H2, <- H3]
+        apply (TensorData.HasShape.HasShapeRoot xs xs.length)
+        simp
+      | td_node xs =>
+        cases xs with
+        | ncons x xs =>
+          intro H2
+          have H3 := (hasShapeNode a H2)
+          revert H3
```
jtristan, line 126: Why do you use this pattern: have ... revert ? If you revert because you want to rewrite, you could also do rw [...] at H3

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+          rw[and_imp]
+          unfold NonEmptyList.length NonEmptyList.all
+          simp
+          intro H3 H4 H5
+          have H6 := (TensorData.HasShape.HasShapeNode (NonEmptyList.ncons x xs) cs)
+          revert H6
+          unfold NonEmptyList.length NonEmptyList.toList
```
jtristan, line 133: you could also do simp [NonEmptyList.length, NonEmptyList.toList] to combine with the following simp

TensorLib/NonEmptyList.lean
```diff
@@ -0,0 +1,25 @@
+namespace TensorLib
+
+inductive NonEmptyList (a: Type) where
```
jtristan, line 3: If we have a single constructor, why not a record?

TensorLib/NonEmptyList.lean
```diff
@@ -0,0 +1,25 @@
+namespace TensorLib
+
```
jtristan, line 2: Use a section?

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+We intend to have a more compact and faster-to-access representation
+in addition to this one in the future for execution.
+-/
+
+-- I tried making this implicit with {} but I needed to add @ everwhere
+-- and give it the argument anyway. Maybe there is a way to infer at least
+-- within a section
+variable (a: Type)
+
+inductive TensorData where
```
jtristan, line 25: Will probably be useful eventually to derive useful instantiations such as Repr

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+| td_root (xs: List a)
+| td_node (xs: NonEmptyList TensorData)
+
+instance : Inhabited (TensorData a) where
+  default := .td_root []
+
+def td_node! (xs: List (TensorData a)): TensorData a :=
```
jtristan, line 32: We need to make a decision. Do we allow ourselves to use "!" functions? I think it may be a mistake.

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+  match x, shape with
+  | .td_root y, List.cons n .nil => List.length y == n
+  | .td_node ys, List.cons n shape' =>
+      ys.length == n &&
+      -- `attach` is used in proofs below, e.g. HashShapeOk
+      List.all ys.toList.attach (fun ⟨ y, _ ⟩ => hasShape y shape')
+  | _, _ => false
+
+lemma hasShapeNil (x: TensorData a): ! (x.hasShape a .nil) := by
+  unfold TensorData.hasShape
```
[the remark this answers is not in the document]
jtristan: Note that for stability purposes, we may (or may not!) want to use simp? and copy the full list of rewrites

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+  simp
+
+lemma hasShapeRoot (xs: List a)
+  (H: (TensorData.td_root xs).hasShape a (.cons c cs)) :
+  cs = .nil ∧ List.length xs = c := by
+  revert H
```
jtristan, line 62: strange revert. You probably need to learn to manipulate formulas in the context directly

TensorLib/TensorData.lean
```diff
@@ -0,0 +1,226 @@
...
+        rw[H2, <- H3]
+        apply (TensorData.HasShape.HasShapeRoot xs xs.length)
+        simp
+      | td_node xs =>
+        cases xs with
+        | ncons x xs =>
+          intro H2
+          have H3 := (hasShapeNode a H2)
+          revert H3
+          rw[and_imp]
```
jtristan, line 127: space between rw and [

TensorLib/TensorElement.lean
```diff
@@ -0,0 +1,26 @@
...
+Authors: John Tristan, Paul Govereau, Sean McLaughlin
+-/
+
+namespace TensorLib
+
+-- Since one goal of TensorLib is to support ML frameworks and languages
+-- like Triton and NKI, we will need to support a menagerie of data types.
+-- Triton and NKI supports signed and unsigned ints (8/16/32/64) and
+-- several floating point types and sizes (float, bfloat, tfloat).
+class TensorElement (a: Type) [Add a][Sub a][Mul a][Neg a] /- etc. -/ where
```
jtristan, line 13: use section, variables?

lakefile.lean
```diff
@@ -10,3 +10,6 @@ lean_lib «TensorLib» where
 @[default_target]
 lean_exe "tensorlib" where
   root := `Main
+
+require mathlib from git
+  "https://github.com/leanprover-community/mathlib4"
```
jtristan, line 15: Specify tag corresponding to the version of lean we're using

TensorLib/TensorElement.lean
seanmcl: Tried this. For some reason Lean doesn't abstract away the type variable in a section. E.g. in 
```
section TensorElement

variable (a: Type)
variable [Add a][Sub a][Mul a][Neg a]

class TensorElement where

end TensorElement

instance : TensorElement Int where
```
the instance doesn't typecheck. Indeed, when I `#check TensorElement` I get `Type` rather than `Type -> instances... -> Type` which is what I expected. Going to leave it as is for now until I figure out how sections work with classes.

lakefile.lean
seanmcl: Good idea. Done.

TensorLib/TensorData.lean
seanmcl: Done throughout.

TensorLib/TensorData.lean
seanmcl: This was just to make the examples that follow without `do` notation. I'll make it private. I'm not actually sure what to do with the examples. Seems like they should be tests or something. Will look what other people do.

TensorLib/TensorData.lean
seanmcl: Yes, implemented `add` yesterday and wanted to print a tensor. That led me to figure out I needed to derive `Repr`. Any others you recommend?

TensorLib/NonEmptyList.lean
seanmcl: I don't grok the rules yet. I still need to type `NonEmptyList a` everywhere even with a section defining `variable (a: Type)`. So nothing was gained when I tried this. What am I missing?

TensorLib/NonEmptyList.lean
seanmcl: Way better.

TensorLib/TensorData.lean
seanmcl: cool

TensorLib/TensorData.lean
seanmcl: Yeah that's better.

TensorLib/NonEmptyList.lean
seanmcl: K, did this and things are slowly getting better

TensorLib/TensorData.lean
seanmcl: Tried a few variations with just messing with H, and none were so succinct. Similar to the other cases below. e.g. when I revert A \/ B, and simp, I can intros A and B separately.

seanmcl:
Moved to https://github.com/leanprover/TensorLib/pull/2 for organizational reasons

PR #2, opened 2024-11-01
Add some basic data and functions

seanmcl:
We've been experimenting with

1. representing tensor data
2. how to define functions
  - naive
  - tail recursive
  - mutually recursive
  - CPS

This will probably be replaced with the numpy flat array representation, but is an interesting experiment nonetheless.


seanmcl:
This is largely obsolete.

PR #4, opened 2024-11-07
Add broadcasting

seanmcl:
This also adds NonEmptyList as a new type for shapes, since shapes can't be empty.


TensorLib/NonEmptyList.lean
```diff
@@ -0,0 +1,118 @@
+import Mathlib.Tactic
```
jtristan, line 1: copyright goes first

TensorLib/NonEmptyList.lean
```diff
@@ -0,0 +1,118 @@
...
+
+namespace NonEmptyList
+
+def toList (x: NonEmptyList a): List a := x.hd :: x.tl
+
+-- Use the nice list bracket concrete syntax
+instance NelRepr [Repr a] : Repr (NonEmptyList a) where
+  reprPrec x n := reprPrec x.toList n
+
+def length (x: NonEmptyList a): Nat := 1 + x.tl.length
```
jtristan, line 25: Is it on purpose that you're not defining a dot notation?

TensorLib/NonEmptyList.lean
```diff
@@ -0,0 +1,118 @@
...
+
+@[simp]
+def append (x : NonEmptyList a) (y : NonEmptyList a) : NonEmptyList a :=
+  NonEmptyList.mk x.hd (x.tl ++ y.toList)
+
```
jtristan, line 30: Is it time to start using docstrings to document all this code?

TensorLib/NonEmptyList.lean
```diff
@@ -0,0 +1,118 @@
...
+  hAppend := appendListR
+
+theorem appendListRLength (x : NonEmptyList a) (y : List a) : (appendListR x y).length = x.length + y.length := by
+  simp [NonEmptyList.length]
+  linarith
+
+theorem hAppendListRLength (x : NonEmptyList a) (y : List a) : (x ++ y).length = x.length + y.length := by
+  apply appendListRLength
+
+def all (x: NonEmptyList a) (P: a -> Bool): Bool :=
```
jtristan, line 71: Could we use coercions to simplify our code?

TensorLib/NumpyRepr.lean
```diff
@@ -100,16 +113,123 @@ instance ByteArrayRepr : Repr ByteArray where
 structure NumpyRepr where
   header : NumpyHeader
   data : ByteArray
-  startIndex : Nat -- Pointer to the first byte of ndarray data
+  startIndex : Nat -- Pointer to the first byte of ndarray data, called `offset` in numpy
```
jtristan, line 116: Why are these not part of the header?

TensorLib/NonEmptyList.lean
seanmcl: Not sure. What do you have in mind?

TensorLib/NonEmptyList.lean
seanmcl: Do you mean converting `NonEmptyList a` to `List a` and using `List` operations instead of duplicating the functionality?

TensorLib/NumpyRepr.lean
seanmcl: I kept the header to what is actually there in the bytes, and added inferred/utility fields here. I'll document this better. To some extent it's arbitrary, though I think there will be multiple metadata objects pointing to the same underlying data so the easier we make that the better.

TensorLib/NumpyRepr.lean
```diff
@@ -100,16 +113,123 @@ instance ByteArrayRepr : Repr ByteArray where
...
+def size (x : NumpyRepr) : ℕ := x.shape.count
+
+def itemsize (x : NumpyRepr) : ℕ := x.header.descr.bytes
+
+def nbytes (x : NumpyRepr) : ℕ := x.itemsize * x.size
+
+section Shape
+
+def M a := Except String a
+variable [Monad M]
```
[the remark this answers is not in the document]
seanmcl: Yeah, this is vestigial. Removed.

TensorLib/NonEmptyList.lean
seanmcl: Resolved offline.

TensorLib/NonEmptyList.lean
govereau: I think he means this: https://leanprover-community.github.io/contribute/doc.html. ?

govereau: Probably we should setup doc-gen4 in the TenCert parent repo, and generate all docs there ?

TensorLib/NumpyRepr.lean
seanmcl: I documented this better.

TensorLib/NonEmptyList.lean
seanmcl: Wow, coercions made the example code a lot nicer. Sold.

PR #5, opened 2024-11-14
Add writing numpy tensors back to disk

seanmcl:
This is needed because while we will generally mutate the data in place, we will need to write new headers, e.g. for the potentially changed shape and row/column ordering. Moreover, for tensors we create from scratch, we will not in general have bits representing the header.

Tested by round-tripping in the python repl.

Unit tests are forthcoming.

### DIFF
--- a/TensorLib/NumpyRepr.lean
+++ b/TensorLib/NumpyRepr.lean
@@ -49,7 +49,7 @@ byte order: https://numpy.org/doc/2.1/reference/generated/numpy.dtype.byteorder.
 '<' is little endian, '|' is "not applicable" Not sure why bool showed up
 as both.
 -/
-def fromString (s: String) : Except String NumpyDtype := match s with
+def fromString (s : String) : Except String NumpyDtype := match s with
 | "<b1" | "|b1"=> .ok bool
 | "|i1" => .ok int8
 | "<i2" => .ok int16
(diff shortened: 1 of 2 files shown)


TensorLib/NumpyRepr.lean
```diff
@@ -514,5 +526,57 @@ def parseFile (path: System.FilePath) : IO (Except String NumpyRepr) := do
 
 end Parse
 
+section Save
+
+private def pushList (a : ByteArray) (xs : List UInt8) : ByteArray := xs.foldl (fun a b => a.push b) a
```
govereau, line 531: Is it more efficient to use ByteArray.append, List.toArray ?

TensorLib/NumpyRepr.lean
```diff
@@ -514,5 +526,57 @@ def parseFile (path: System.FilePath) : IO (Except String NumpyRepr) := do
...
+
+private def pushChars (a : ByteArray) (xs : List Char) : ByteArray := xs.foldl (fun a b => a.push (b.toUInt8)) a
+
+private def pushString (a : ByteArray) (xs : String) : ByteArray := pushChars a xs.toList
```
govereau, line 535: Similar comment: String.toUTF8?

govereau (approved):

TensorLib/NumpyRepr.lean
seanmcl: Probably.  Will swap.

TensorLib/NumpyRepr.lean
seanmcl: ah didn't know about that one

PR #10, opened 2024-11-20
Allow empty shapes

seanmcl:
We made a mistaken assumption that None is the only element that can yield an empty shape. Array scalars like `np.array(3)` also have empty shapes, and are used in various ways in np operators.


govereau (approved):
Looks good. Not needing NonEmptyList seems like a good side benefit.

PR #17, opened 2025-01-02
Npy file support

seanmcl:
NumPy uses a custom binary format to store ndarrays. This file implements that format.

### DIFF
--- a/TensorLib/Npy.lean
+++ b/TensorLib/Npy.lean
@@ -1,0 +1,410 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan, Paul Govereau, Sean McLaughlin
+-/
+
+import TensorLib.Common
+import TensorLib.Dtype
+
+/-!
+We largely duplicate the NumPy representation of tensors.
+
+The binary format is described here: https://numpy.org/doc/stable/reference/generated/numpy.lib.format.html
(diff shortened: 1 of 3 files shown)


TensorLib/Npy.lean
```diff
@@ -0,0 +1,397 @@
...
+| '>' => .ok bigEndian
+| '|' => .ok notApplicable
+| _ => .error s!"can't parse byte order: {c}"
+
+end ByteOrder
+
+structure Dtype where
+  name : TensorLib.Dtype.Name
+  order : ByteOrder
+deriving BEq, Repr, Inhabited
```
govereau, line 59: Isn't this defined already in `Dtype.lean`?

TensorLib/Npy.lean
```diff
@@ -0,0 +1,397 @@
...
+
+https://numpy.org/doc/stable/reference/generated/numpy.lib.format.html#format-version-1-0
+-/
+structure Header where
+  major : UInt8 := 1
+  minor : UInt8 := 0
+  descr : Dtype
+  dataOrder : DataOrder := DataOrder.C
+  shape : Shape
+  deriving Repr, Inhabited
```
govereau, line 120: Side comment, I have a "tensor type" in NKL, which is basically this structure.  However, I need BEq.

govereau (approved):

TensorLib/Npy.lean
seanmcl: Yeah, but this one supports "native" which I want to avoid but need to support parsing. There is some duplicate constructors but it's not too awful I think. I thought about trying to use Lean to get the byte ordering of the processor but decided against it. We can revisit later?

TensorLib/Npy.lean
seanmcl: Let's not use `Header` for that. This is supposed to be an annoying implementation detail. You don't need the version number for example.

PR #18, opened 2025-01-02
Broadcasting

seanmcl:
This is some utility functions related to broadcasting.


govereau (approved):
Nice, I can use this directly in NKL.

PR #21, opened 2025-01-02
Basic indexing

seanmcl:
"Basic" indexing means indexing with a constant or slice on all the dimensions. (Advanced indexing is for things like indexing by other arrays).

https://numpy.org/doc/stable/user/basics.indexing.html

### DIFF
--- a/TensorLib/Index.lean
+++ b/TensorLib/Index.lean
@@ -1,0 +1,388 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan, Paul Govereau, Sean McLaughlin
+-/
+
+import TensorLib.Common
+import TensorLib.Tensor
+import TensorLib.Slice
+import TensorLib.Npy
+
+/-
+Theorems to prove (taken from NumPy docs):
+
+1. Basic slicing with more than one non-: entry in the slicing tuple,
+   acts like repeated application of slicing using a single non-: entry,
+   where the non-: entries are successively taken (with all other non-: entries replaced by :).
+   Thus, x[ind1, ..., ind2,:] acts like x[ind1][..., ind2, :] under basic slicing.
+
+2. Advanced indices always are broadcast and iterated as one:
+result[i_1, ..., i_M] == x[ind_1[i_1, ..., i_M], ind_2[i_1, ..., i_M],
+                           ..., ind_N[i_1, ..., i_M]]
+3. ...TODO...
+-/
+
+namespace TensorLib
+namespace Index
+
+-- Parsed form. We will simplify some of the redundancy here to yield an Index.Basic
+-- In NumPy there may only be a single ellipsis present.
+inductive NumpyItem where
+| int (n : Int)
+| slice (slice : Slice)
+| ellipsis
+| newaxis
+deriving BEq, Repr
+
+abbrev NumpyBasic := List NumpyItem
+
+-- Slices do not subsume nat indexing. Slices always return an array
+-- with the same dimension as the input, while a nat index always reduces the
+-- dimensions by one.
+-- x
+-- array([1, 2, 3, 4, 5, 6])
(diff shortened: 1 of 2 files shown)


TensorLib/Index.lean
```diff
@@ -0,0 +1,352 @@
...
+-- Slices do not subsume nat indexing. Slices always return an array
+-- with the same shape as the input, while a nat index always reduces the shape by one dimension.
+-- x
+-- array([1, 2, 3, 4, 5, 6])
+--
+-- x[0], x[0:1:1]
+-- (np.int64(1), array([1]))
+--
+-- x[0].shape, x[0:1:1].shape
+-- ((), (1,))
```
govereau, line 49: Do you mean same "dimension" not "shape"?
I think the shape of x is `(6,)`, which isn't the same as `(1,)`

TensorLib/Index.lean
```diff
@@ -0,0 +1,352 @@
...
+    | .newaxis :: items, dim :: shape => do
+      let (basic, shape) <- loop items shape
+      let slice := Slice.Iter.make Slice.all dim
+      .ok (.slice slice :: basic, dim :: shape)
+    | .ellipsis :: items, dim :: shape => do
+      if items.length == 1 + shape.length then
+        loop items (dim :: shape)
+      else  -- there are fewer items than the shape
+        loop (.ellipsis :: items) shape
+    | _, _ => .error "impossible"
```
govereau, line 91: I was confused here for a second, until I noticed the length comparison at the start.
Is this case checking the same thing as the line on 63? (maybe a comment would be helpful)

TensorLib/Index.lean
```diff
@@ -0,0 +1,352 @@
...
+        (n :: ns, basic)
+
+def next (iter : BasicIter) : Option (List Nat × BasicIter) :=
+  if iter.done then none else
+  let (ns, basic) := nextWithCarry iter.basic false
+  let ns := ns.reverse
+  let basic := match basic with
+  | none => { iter with done := true } -- All slice iterators are maxxed out in this case. Should we check this?
+  | some basic => { iter with basic }
+  some (ns, basic)
```
govereau, line 197: The algorithm here has a lot of finicky details, but it is still pretty easy to read.
Nice job!

TensorLib/Index.lean
```diff
@@ -0,0 +1,352 @@
...
+    let mut iter : BasicIter := iter
+    let mut res := x
+    for _ in [0:iter.size] do
+      match iter.next with
+      | .none => break
+      | .some (ns, iter') =>
+        iter := iter'
+        match <- f ns res with
+        | .yield k
+        | .done k => res := k
```
govereau, line 210: May have fixed already, the issue with `break` or `return` mentioned in prior reviews.

TensorLib/Index.lean
```diff
@@ -0,0 +1,352 @@
...
+  for dimIndex in iter do
+    let posn := arr.dimIndexToPosition dimIndex
+    for j in [0:itemsize] do
+      let b := arr.data.get! (posn + j)
+      data := data.push b
+  return {
+     dtype := arr.dtype,
+     shape := newShape,
+     data
+  }
```
govereau, line 228: Cool, I wish I had read this function first. :)

govereau (approved):

govereau (approved):

TensorLib/Index.lean
seanmcl: Nice comment. I lifted the error case to the top so it's more obvious
```
    | [], shape => .ok (shapeToSliceIters shape, shape)
    | _, [] => .error "impossible" -- We checked above that there are at least as many dims as items
```

TensorLib/Index.lean
seanmcl: Fixed and added similar tests.

TensorLib/Index.lean
seanmcl: Ha. Yes I kind of miss out-of-order definitions for this reason. Doing lots of gritty work so the top level stuff is pretty.

PR #25, opened 2025-01-07
Advanced indexing

seanmcl:
We implement a common version of advanced indexing, where all arguments are (broadcastable to) int arrays.

### DIFF
--- a/TensorLib/Common.lean
+++ b/TensorLib/Common.lean
@@ -5,7 +5,6 @@ Authors: Jean-Baptiste Tristan, Paul Govereau, Sean McLaughlin
 -/
 
 import Std.Tactic.BVDecide
-
 namespace TensorLib
 
 --! The error monad for TensorLib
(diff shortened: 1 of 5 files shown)


TensorLib/Common.lean
```diff
@@ -223,6 +263,21 @@ def allDimIndices (shape : Shape) : List DimIndex := Id.run do
...
+-- dimension in question. It only supports `-n` to `n`.
+def intIndexToDimIndex (shape : Shape) (index : List Int) : Err DimIndex := do
+  if shape.ndim != index.length then .error "intsToDimIndex length mismatch" else
+  let conv (dim : Nat) (ind : Int) : Err Nat :=
+    if 0 <= ind then
+      if ind < dim then .ok ind.toNat
+      else .error "index out of bounds"
+    else if ind < -dim then .error "index out of bounds"
+    else .ok (dim + ind).toNat
+  (shape.val.zip index).traverse (fun (dim, ind) => conv dim ind)
```
govereau, line 276: I think this is why you have batteries, for `traverse`. Why is this not doable with `mapM`?

govereau (approved):
Looks good. I think it is safe (for now) to not support mixing, I think NKI doesn't handle those cases either.

TensorLib/Common.lean
seanmcl: Ah, `mapM` is better. Changed throughout the codebase. I usually just reach for `traverse` since it works in more cases, but here all we have are monads so we are good either way.

(Still use Batteries.Data.List for `toChunks` in one place, but that's it. We could implement this ourselves if we want to ever drop the Batteries/Aesop dependency).

PR #29, opened 2025-01-21
Universal functions

seanmcl:
NumPy uses an abstraction of functions called [Ufunc](https://numpy.org/doc/2.1/reference/ufuncs.html). It has a lot of widgets in NumPy, but for now we just support lifting element-wise operations and broadcasting.

### DIFF
--- a/TensorLib/Ufunc.lean
+++ b/TensorLib/Ufunc.lean
@@ -1,0 +1,187 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan, Paul Govereau, Sean McLaughlin
+-/
+
+import TensorLib.Common
+import TensorLib.Tensor
+import TensorLib.Broadcast
+
+/-!
+Universal functions: https://numpy.org/doc/stable/reference/ufuncs.html
+-/
+
+namespace TensorLib
+namespace Tensor
+namespace Ufunc
+
+private def binop (a : Type) [Element a] (x y : Tensor) (op : a -> a -> Err a) : Err Tensor :=
+  match Broadcast.broadcast { left := x.shape, right := y.shape } with
+  | .none => .error s!"Can't broadcast shapes ${x.shape} with {y.shape}"
+  | .some shape =>
+    if x.dtype != y.dtype then .error s!"Casting between dtypes is not implemented yet: {repr x.dtype} <> {repr y.dtype}" else
+    do
+      let mut arr := Tensor.empty x.dtype shape
+      let iter := DimsIter.make shape
+      for idx in iter do
+        let v <- Element.getDimIndex x idx
+        let w <- Element.getDimIndex y idx
+        let k <- op v w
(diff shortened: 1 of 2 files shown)


TensorLib/Ufunc.lean
```diff
@@ -0,0 +1,180 @@
...
+private def uniq [BEq a] (xs : List a) : Bool := match xs with
+| [] | [_] => true
+| x1 :: x2 :: xs => x1 != x2 && uniq (x2 :: xs)
+
+def sum (a : Type) [Add a] [Zero a] [Element a] (arr : Tensor) (axes : Option (List Nat)) : Err Tensor :=
+  match axes with
+  | .none => sum0 a arr
+  | .some axes =>
+  if !(uniq axes) then .error "Duplicate axis elements" else
+  let axes := (List.mergeSort axes).reverse
```
govereau, line 90: Ah here is the sort, but maybe in the wrong place?

govereau (changes requested):

TensorLib/Ufunc.lean
```diff
@@ -0,0 +1,180 @@
...
+  binop a x y (fun x y => .ok (x + y))
+
+def sub (a : Type) [Sub a] [Element a] (x y : Tensor) : Err Tensor :=
+  binop a x y (fun x y => .ok (x - y))
+
+def mul (a : Type) [Mul a] [Element a] (x y : Tensor) : Err Tensor :=
+  binop a x y (fun x y => .ok (x * y))
+
+def div (a : Type) [Div a] [Element a] (x y : Tensor) : Err Tensor :=
+  binop a x y (fun x y => .ok (x / y))
```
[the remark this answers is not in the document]
seanmcl: Numpy just gives inf or nan in this case. E.g.
```
#np.full(5, 1, dtype='int8') / np.arange(5)
<ipython-input-26-4ea893b50512>:1: RuntimeWarning: divide by zero encountered in divide
  np.full(5, 1, dtype='int8') / np.arange(5)
array([       inf, 1.        , 0.5       , 0.33333333, 0.25      ])
```

Seems like the correct behavior to me. E.g. if we fail on nans we will have quite a different behavior than the current numpy and neuron compiler.

TensorLib/Ufunc.lean
```diff
@@ -0,0 +1,180 @@
...
+      for i in [0:dim] do
+        let index' := index.insertIdx axis i
+        let v : a <- Element.getDimIndex arr index'
+        acc := acc + v
+      res <- Element.setDimIndex res index acc
+    return res
+
+private def uniq [BEq a] (xs : List a) : Bool := match xs with
+| [] | [_] => true
+| x1 :: x2 :: xs => x1 != x2 && uniq (x2 :: xs)
```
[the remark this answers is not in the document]
seanmcl: Good catch, fixed.

TensorLib/Ufunc.lean
seanmcl: yeah, thanks

seanmcl: Added a test.

govereau (approved):

PR #32, opened 2025-02-16
chore: Remove wrong index compatibility check

seanmcl:
The best-effort check we're doing to ensure an index is "compatible" with a shape is wrong headed. For example, if we're indexing a single element from an array, the output shape is [] but we're comparing it to the list of nat indices. Instead of trying to improve the notion, I just removed it and let the indexing itself fail if there's an incorrect index.

### DIFF
--- a/TensorLib/Index.lean
+++ b/TensorLib/Index.lean
@@ -144,21 +144,7 @@ deriving Inhabited, Repr
 
 namespace BasicIter
 
--- An iterator is compatible with a shape if all of the shape's dimensions are
--- larger than the constant indices and equal to the iterator indices.
-private def compatibleWith (iter : BasicIter) (shape : List Nat) : Bool :=
-  let rec loop := fun
-  | [], [] => true
-  | .nat n :: basic, dim :: shape => n < dim && loop basic shape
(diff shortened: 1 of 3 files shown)


TensorLib/Index.lean
```diff
@@ -144,21 +144,7 @@ deriving Inhabited, Repr
...
-  | .slice iter :: basic, dim :: shape => dim <= iter.dim && loop basic shape
-  | _, _ => false
-  loop iter.basic shape
-
-def make (shape : Shape) (basic : Basic) : Err BasicIter :=
-  let iter := BasicIter.mk basic.reverse false
-  if ! compatibleWith iter shape.val.reverse
-  then .error s!"shape/index mismatch: {shape.val} : {basic.repr 100}"
-  else .ok iter
+def make (basic : Basic) : BasicIter :=  BasicIter.mk basic.reverse false
```
govereau, line 147: How come you don't need the compatibility check anymore?

govereau (approved):

TensorLib/Index.lean
govereau: Oh, nvm. I see in the commit description.

PR #33, opened 2025-02-16
feat: index assignment

seanmcl:
Implements assigning an indexed tensor to another tensor of the same shape (after indexing). For example,

    # x = np.arange(20).reshape(4,5)
    # x[1:3, 1:4] = np.array([[40, 50, 60], [70, 80, 90]]); x
    array([[ 0,  1,  2,  3,  4],
           [ 5, 40, 50, 60,  9],
           [10, 70, 80, 90, 14],
           [15, 16, 17, 18, 19]])

### DIFF
--- a/TensorLib/Index.lean
+++ b/TensorLib/Index.lean
@@ -297,6 +297,38 @@ def apply (index : NumpyBasic) (arr : Tensor) : Err (Tensor × Bool) := do
     return (res, false)
 
 /-
+`arr[index] = v`
+
+Since the dtypes need to be equal, for now, Element here is awkward. All we need it the byte size, which is in the Tensor
+Refactoring would be good here.
+
+Note that v is broadcast to arr[index].shape, but arr[index].shape is not broadcast to v, which wouldn't make sense.
+E.g.
+
+# x = np.arange(6).reshape(2, 3)
+# x[1,:] = np.array([5,6,7])
+
+is ok but
+
+# x[1,:] = np.array([[1,2,3], [4,5,6]])
+
+wouldn't make any sense, even though the shapes (1, 3) and (2, 3) are broadcastable.
+-/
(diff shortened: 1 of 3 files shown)


TensorLib/Index.lean
```diff
@@ -296,6 +296,27 @@ def apply (index : NumpyBasic) (arr : Tensor) : Err (Tensor × Bool) := do
...
+-- Refactoring would be good here.
+def assign (a : Type) [typ : Tensor.Element a] (arr : Tensor) (index : NumpyBasic) (v : Tensor) : Err Tensor := do
+  if arr.dtype != v.dtype || typ.dtype != arr.dtype then .error "Type mismatch" else
+  let (basic, shape) <- toBasic index arr.shape
+  match Broadcast.broadcast (Broadcast.mk shape v.shape) with
+  | .none => .error "Can't broadcast shapes"
+  | .some vShape =>
+  let v <- v.reshape vShape
+  let aIter := BasicIter.make basic
+  let mut vIter := DimsIter.make shape
```
govereau, line 310: I think it doesn't matter, because of the broadcast, but I was expecting to see `vShape` here.

govereau (approved):
Very nice.

TensorLib/Index.lean
seanmcl: No, this points a bug. But I think it can't be fixed with just vShape. For example, I didn't test broadcasting on the right

```
# x = np.arange(20).reshape(4,5)
# x[1:3, 1:4] = np.array([40, 50, 60]); x
array([[ 0,  1,  2,  3,  4],
       [ 5, 40, 50, 60,  9],
       [10, 40, 50, 60, 14],
       [15, 16, 17, 18, 19]])
```

This example crashes with the current implementation. Will work on it, thanks for noticing!

PR #38, opened 2025-02-22
feat: matmul

seanmcl:
This is pretty close to the NumPy version. We don't handle 1D arrays yet, but N-D seems to work. The method is inefficient, with a broadcast/reshape/reshape, but is conceptually simple.


govereau (approved):
Looks really nice.

PR #39, opened 2025-02-25
chore: Add Dtype bits and arithmetic logic

seanmcl:
Dtype will soon replace Element as the method of doing array indexing and arithmetic. This PR adds some methods to that end.

### DIFF
--- a/TensorLib/Npy.lean
+++ b/TensorLib/Npy.lean
@@ -48,11 +48,11 @@ deriving BEq, Repr, Inhabited
 
 namespace ByteOrder
 
-def toByteOrder (x : ByteOrder) (default : TensorLib.ByteOrder): TensorLib.ByteOrder := match x with
-| .native => default
-| .littleEndian => .littleEndian
-| .bigEndian => .bigEndian
-| .notApplicable => .oneByte
+def toByteOrder (x : ByteOrder) : Option TensorLib.ByteOrder := match x with
+| .native => none
+| .littleEndian => some .littleEndian
+| .bigEndian => some .bigEndian
+| .notApplicable => some .oneByte
 
 def toChar (x : ByteOrder) := match x with
 | native => '='
@@ -150,7 +150,15 @@ structure Ndarray where
   startIndex : Nat -- First byte of non-header data
   deriving Repr, Inhabited
 
-def Ndarray.nbytes (x : Ndarray) : Nat := x.header.descr.itemsize * x.header.shape.count
+namespace Ndarray
+
+def nbytes (x : Ndarray) : Nat := x.header.descr.itemsize * x.header.shape.count
+
+def dtype (arr : Ndarray) : Dtype := arr.header.descr
+
+def order (arr : Ndarray) : ByteOrder := arr.dtype.order
+
+end Ndarray
 
 section Parse
 


TensorLib/Dtype.lean
```diff
@@ -28,11 +28,13 @@ deriving BEq, Repr, Inhabited
 namespace Name
 
 instance : ToString Name where
-  toString x := (repr x).pretty
+  toString x := ((repr x).pretty.splitOn ".").getLast!
```
govereau, line 31: Maybe:
```
if .str _ last := x then last else x.toString
```

TensorLib/Dtype.lean
```diff
@@ -61,21 +63,258 @@ deriving BEq, Repr, Inhabited
...
+    byteArrayOfNat dtype (x + y)
+  | .int8 | .int16| .int32 | .int64 => do
+    let x <- byteArrayToInt dtype x
+    let y <- byteArrayToInt dtype y
+    byteArrayOfInt dtype (x + y)
+  | .float64 => do
+    let x <- byteArrayToFloat dtype x
+    let y <- byteArrayToFloat dtype y
+    byteArrayOfFloat dtype (x + y)
+  | .bool | .float16 | .float32 => .error s!"`add` not supported at type ${dtype.name}"
```
govereau, line 261: Why not bool? Numpy seems to be OK with it.

TensorLib/Dtype.lean
```diff
@@ -61,21 +63,258 @@ deriving BEq, Repr, Inhabited
...
+
+def sub (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
+  if dtype.itemsize != x.size || dtype.itemsize != y.size then .error "sub: byte size mismatch" else
+  match dtype.name with
+  | .uint8 | .uint16 | .uint32 | .uint64 => do
+    let x <- byteArrayToNat dtype x
+    let y <- byteArrayToNat dtype y
```
govereau, line 268: Is this right? In lean `4-3=0` at type nat, but I think we want `4-3=0xfff...`?

govereau (commented):
Some minor questions.

TensorLib/Dtype.lean
seanmcl: Sorry, not following. repr of this thing is `"TensorLib.Dtype.Name.uint8"` (or sth similar). I just wanted to drop everything but the last bit. What is `.str` in your example?

TensorLib/Dtype.lean
seanmcl: 3-4?

Oh hmm, you want to implement underflow. OK, that is an option. One thing I like about this wall of code is that it's clear and in one place what all these finicky details are doing.

TensorLib/Dtype.lean
seanmcl: Funny, + is supported but - isn't. Sure I can implement bool. This stuff to some extent gets mixed up with casting (e.g. `np.astype`) which is coming in my next PR, but I'll try to support everything we can for now. The tricky part of numpy is that it's not always easy to tell when a cast is happening, and what that cast is

TensorLib/Dtype.lean
seanmcl: I guess if we stick with this we should do overflow also at uint types. I'm not sure how I feel about them. Wouldn't overflow and underflow almost always result in a computation you didn't mean.

seanmcl:
Sorry, merged this by accident when trying to merge the npy file change that was stacked on top of this. Not sure exactly how I messed up, but will create a new PR with the changes proposed.

TensorLib/Dtype.lean
govereau: The `Name` "TensorLib.Dtype.Name.uint8" is equal to:
```
.str (.str (.str (.str .anonymous "TensorLib") "Dtype") "Name") "uint8"
```
So, `pretty` followed by `splitOn "."` can be done by just matching on the `Name`.

TensorLib/Dtype.lean
govereau: Yeah, I don't love it either: the underflow is messy to write, and probably not what the user wanted. I will let you make the call :)

TensorLib/Dtype.lean
seanmcl: Rewritten

TensorLib/Dtype.lean
seanmcl: I'm going to incorporate this into a bigger change where there are both checked and unchecked versions of the constructors and arithmetic operators.

PR #41, opened 2025-02-25
chore: Remove Element type class. Use Dtype instead.

seanmcl:
This pervasive change replaces the `Element` type class for doing indexing and arithmetic with Dtype. I was always unsatisfied with having both Element and Dtype, but didn't know how to do type-based Lean arithmetic without casting to Lean types. This added conceptual complexity; the dtype already told you exactly what the representation in Lean should be. Now we do all arithmetic and indexing via ByteArray, using the dtype to convert between the raw bytes and the corresponding arithmetic operation. This makes the code a lot more clear, with less polymorphism and more direct pattern matching. I think the directness is a win. In a later PR we will see `numpy.astype`-style casting, and the dtype approach makes this much more straightforward than the type class approach.


TensorLib/Tensor.lean
```diff
@@ -311,131 +312,107 @@ def broadcast (arr1 : Tensor) (arr2 : Tensor) : Err (Tensor × Tensor) :=
...
+  let posn := n * size
+  { arr with data := v.copySlice 0 arr.data posn size }
+
+#guard
+  let tp :=  Dtype.uint8
+  let arr := arange! tp 4
+  let arr := setPosition arr 0 (tp.byteArrayOfNat! 7)
+  arr.data.data == #[7, 1, 2, 3]
+
+def ofNatList (dtype : Dtype) (ns : List Nat) : Err Tensor := do
```
govereau, line 357: I wonder, would it make sense to tie the dtypes to native types with an associated type?
```
class DtypeOf a where
  dtype : Dtype
```
Then, maybe `ofNatList` and `ofIntList` could be:
```
def ofList [inst:DtypeOf a] : List a -> Err Tensor := do
  let dtype := inst.dtype
  ...
```
??

TensorLib/Test.lean
```diff
@@ -11,47 +11,28 @@ namespace TensorLib
...
 -- Sketchy perhaps, but seems to work for testing
-private def ioBool (x : IO Bool) : Bool := match x.run () with
-| .error _ _ => false
-| .ok b _ => b
-
-#guard ioBool (testTensorElementBV 16 "uint16")
-#guard ! ioBool (testTensorElementBV 32 "uint16")
-#guard ioBool (testTensorElementBV 32 "uint32")
-#guard ! ioBool (testTensorElementBV 32 "uint64")
+private def runIo [Inhabited a] (x : IO a) : a := match x.run () with
```
govereau, line 31: Whoa, it took me a minute to figure out how this works. I guess you know that `IO.RealWorld` is `Unit`. Maybe just use `IO.RealWorld`, in case they change it? (the code in prelude says: "TODO: make opaque")

govereau (approved):
Really nice change.
I didn't read carefully all of the code under `#guard`, I am trusting that those tests make sense.

TensorLib/Tensor.lean
```diff
@@ -311,131 +312,107 @@ def broadcast (arr1 : Tensor) (arr2 : Tensor) : Err (Tensor × Tensor) :=
...
-  let itemsize := typ.itemsize
-  if itemsize != x.itemsize then .error "byte size mismatch" else -- TODO: Lift this check out so we only do it once
-  let bytes := typ.toByteArray v
-  let posn := n * itemsize
-  .ok { x with data := bytes.copySlice 0 x.data posn itemsize true }
-
-def ofList (a : Type) [typ : Element a] (xs : List a) : Tensor := Id.run do
-  let arr := Tensor.zeros typ.dtype (Shape.mk [xs.length])
+def getPosition (arr : Tensor) (position : Nat) : ByteArray :=
+  arr.data.copySlice (position * arr.itemsize) ByteArray.empty 0 arr.itemsize
```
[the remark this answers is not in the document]
seanmcl: Not sure. The Lean code doesn't do much of this, e.g.

```
def extract (a : ByteArray) (b e : Nat) : ByteArray :=
  a.copySlice b empty 0 (e - b)
```

could have been a `mkEmpty (e - b)`. I'll change it anyway. Makes it a bit more clear to the reader even what is going on.

TensorLib/Test.lean
seanmcl: Moved this to Main

PR #42, opened 2025-02-26
chore: fix build, address comment from previous PR

seanmcl:
As I was testing testing, I forgot to fix my test. Also responded to an unresolved comment from
https://github.com/leanprover/TensorLib/pull/39


seanmcl:
FWIW, I tried to add main branch protection so I can't accidentally push changes to main without review. I'll try that when this is approved to ensure it is working.

govereau (approved):

PR #48, opened 2025-03-04
chore: Make all arrays little-endian

seanmcl:
Endian-ness is a nice feature in NumPy; it's clear what the data should represent, it handles operations on arrays with different orderings, etc. But it's largely just an optimization. In our code, which is slow to begin with, we do not need this optimization. We remove considerable complexity by forcing all arrays in TensorLib to be little-endian.

One outcome of this is that we will save all .npy files as littleendian and rely on the Python programmer to convert if necessary.


govereau (approved):
Nice. Lots of simplification.

PR #49, opened 2025-03-07
fix: Fix bug in Slice.size, prove size d <= d

seanmcl:
Bug was found while working on the proof.


seanmcl (commented):
I removed the dependent match here because it was very hard to prove things about. John Tristan said they never use dependent matches in compcert for this reason.

govereau (approved):
Nice catch. Hooray for theorem proving.

PR #50, opened 2025-03-12
chore: improve iterators

seanmcl:
TensorLib (and soon KLR) are rife with iterators. We were rolling them ad hoc, with a lot of duplicate code. For example, an Mgrid iterator is really just a carry-adder-like list of int iterators. A shape iterator is just a carry-adder-like list of nat iterators. There are some weird ones still like Slice iterators that stick around, but mostly we pulled the generic ones into a type class.

I found a much nicer way to do carry-adder iterators, but it involved a signature change where not all type variables are returned by all type class functions. This means there is some annoying instruction of the missing types in some calls. Overall, though, a massive simplification.


TensorLib/Index.lean
```diff
@@ -264,38 +143,6 @@ private def applyWithCopy (index : NumpyBasic) (arr : Tensor) : Err Tensor := do
...
-/-
-Make an attempt to not copy. For example, a partial index into an array
-should probably just return a view with the same (sub) strides.
-This is currently a feeble attempt, but can get fancier over time.
-We currently handle
-
-1. Non-slice indices (e.g. arr[1][7][-2])
-2. TODO
--/
-def apply (index : NumpyBasic) (arr : Tensor) : Err Tensor := do
```
seanmcl, line 276: I removed this bc it adds complexity, and is incompatible with stuffing nat indicies into singleton slices, which is how I simplified the old "Basic" iterator without needing a "sum" iterator. I may add that if we have another example, but this workaround seemed nice for code simplicity.

TensorLib/Iterator.lean
```diff
@@ -0,0 +1,316 @@
...
+  let iter := (iter1, iter1)
+  Iterator.toList iter == [(0, 0), (1, 1), (2, 2), (3, 3), (4, 4)]
+
+-- A carry-adder-style pair iterator
+instance instPairCarry [inst1 : Iterator i1 v1] [inst2 : Iterator i2 v2] : Iterator (i1 × i2) (v1 × v2) where
+  size := fun (l, r) => inst1.size l * inst2.size r
+  reset := fun (l, r) => (inst1.reset l, inst2.reset r)
+  peek := fun (l, r) => (inst1.peek l, inst2.peek r)
+  next := fun (l, r) =>
+    match inst1.next l with
```
govereau, line 190: This came out nice.

TensorLib/Iterator.lean
```diff
@@ -0,0 +1,316 @@
...
+  next iter := (next' iter.val []).map LEList.mk
+where
+  -- We are searching for an iter to increment. All elements of `accT` are the previous iterators (reversed)
+  -- which were all maxxed out. `accV` is the 0th elements of the corresponding iterators. We assume
+  -- all iterators to the right (i.e. in `iters`) have
+  next' (iters : List iter) (acc : List iter) : Option (List iter) :=
+    match iters with
+    | [] => none
+    | iter :: iters =>
+      match inst.next iter with
```
govereau, line 251: This reminds me a little bit of the "zipper" structures. I wonder if you have `LEList` be two lists: the maxed out and not maxed out, if you could avoid traversing and rebuilding the list on each call to next?

TensorLib/Iterator.lean
```diff
@@ -0,0 +1,316 @@
...
+
+def make (iters : List iter) : BEList iter := BEList.mk iters.reverse
+
+instance instListBE [inst : Iterator iter value] : Iterator (BEList iter) (List value) where
+  size iter := iter.val.foldl (fun acc i => acc * inst.size i) 1
+  reset iter := BEList.mk $ iter.val.map inst.reset
+  peek iter := (iter.val.map inst.peek).reverse
+  next iter := (next' iter.val []).map BEList.mk
+where
+  next' (iters : List iter) (acc : List iter) : Option (List iter) :=
```
govereau, line 286: Is this identical to `instListLE.next'`?

TensorLib/Mgrid.lean
```diff
@@ -33,213 +34,49 @@ array([], dtype=int64)
...
-#guard (sliceSize (Slice.make! (.some (-5)) (.some (-10)) (.some (-1)))) == .some 5
-#guard (sliceSize (Slice.make! (.some (-5)) (.some 10) (.some (-1)))) == .some 0
-#guard (sliceSize (Slice.make! (.some (-5)) (.some 10) (.some 1))) == .some 15
-#guard (sliceSize (Slice.make! (.some (-5)) (.some 10) (.some (-3)))) == .some 0
-#guard (sliceSize (Slice.make! (.some (-5)) (.some 10) (.some 3))) == .some 5
+def sliceToIntIter (slice : Slice) : Err Iterator.IntIter :=
+  let step := slice.step.getD 1
+  match slice.start, slice.stop with
+  | some start, some stop => Iterator.IntIter.make start stop step
+  | _, _ => .error "Can't convert slice to int iterator"
```
govereau, line 77: Should a start of none be the same as a start of 1?

govereau (approved):

TensorLib/Iterator.lean
```diff
@@ -0,0 +1,316 @@
...
+  next : iter -> Option iter
+  peek : iter -> value
+  size : iter -> Nat
+  reset : iter -> iter
+namespace Iterator
+
+-- https://leanprover-community.github.io/archive/stream/270676-lean4/topic/cannot.20find.20synthesization.20order.20for.20instance.html
+set_option synthInstance.checkSynthOrder false
+
+instance forInInstance [Monad m] [inst : Iterator iter value] : ForIn m iter value where
```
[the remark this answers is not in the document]
seanmcl: Yes!

TensorLib/Iterator.lean
```diff
@@ -0,0 +1,316 @@
...
+deriving Repr
+
+instance : Inhabited IntIter where
+  default := IntIter.mk 0 1 1 0 (by simp)
+
+namespace IntIter
+
+def make (start stop step : Int) : Err IntIter :=
+  if H : step == 0 then .error "step can't be 0" else
+  let stepNz : step ≠ 0 := by aesop
```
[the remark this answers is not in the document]
seanmcl: Yes. I like automation rather than spending time on stuff like this in general, but recognize the cost down the road.

TensorLib/Iterator.lean
seanmcl: Thanks. Yes, when I separated the next value itself from bumping the iterator things just fell into place.

TensorLib/Iterator.lean
seanmcl: I was not looking forward to abstracting this, but it enabled me to remove both `where` clauses, so glad for the push!

TensorLib/Iterator.lean
seanmcl: Yeah, intersting idea. Moving the zipper around to where the work is happening is interesting. Added a TODO.

TensorLib/Mgrid.lean
seanmcl: Don't think so. If the step is negative it's not obvious where you want to start. I added a comment.

PR #54, opened 2025-03-18
chore: Remove evals

seanmcl:
One IO #eval seems to be causing problems when used as a library in a different repo. As a matter of tidiness I replaced uses of #eval with #guard/example. I think keeping eval out of our code as a general rule is a good idea, as it tends to just dump output during builds. #guard/#guard_msgs is effective for local testing.

### DIFF
--- a/TensorLib/Test.lean
+++ b/TensorLib/Test.lean
@@ -27,10 +27,6 @@ private def testTensorElementBV (dtype : Dtype) : IO Bool := do
   let expected := (Tensor.arange! dtype 20).reshape! (Shape.mk [5, 4])
   return Tensor.arrayEqual expected arr
 
-#eval do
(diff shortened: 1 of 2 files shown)


TensorLib/Test.lean
```diff
@@ -27,10 +27,6 @@ private def testTensorElementBV (dtype : Dtype) : IO Bool := do
   let expected := (Tensor.arange! dtype 20).reshape! (Shape.mk [5, 4])
   return Tensor.arrayEqual expected arr
 
-#eval do
-  let b <- testTensorElementBV Dtype.uint16
```
seanmcl, line 31: This test runs during `tensorlib test` below, so doesn't need to be evaluated also at build time.

govereau (approved):

PR #56, opened 2025-03-19
chore: improve lock-step iterators

seanmcl:
There are two common pair iterator use cases
1. lock-step
2. carry-adder

This creates a scoped instance for (1) so you
can use `open scoped` to get the syntax we want
for lock-step iterators

    for (i, j) in (iter1, iter2) do ...


seanmcl:
Based on a comment from Paul in https://github.com/leanprover/KLR/pull/81 that code that used two iterators was "pretty dense".

govereau (approved):

PR #60, opened 2025-03-27
feat: Add trig functions

seanmcl:
We get our first rounding error here

### DIFF
--- a/TensorLib/Dtype.lean
+++ b/TensorLib/Dtype.lean
@@ -82,6 +82,10 @@ def isUint (x : Dtype) : Bool := match x with
 
 def isIntLike (x : Dtype) : Bool := x.isInt || x.isUint
 
+def isFloat (x : Dtype) : Bool := match x with
+| .float32 | .float64 => true
+| _ => false
+
 --! Number of bytes used by each element of the given dtype
 def itemsize (x : Dtype) : Nat := match x with
 | float64 | int64 | uint64 => 8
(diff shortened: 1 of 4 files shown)


TensorLib/Dtype.lean
```diff
@@ -437,6 +451,131 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err
...
+    let f <- Float32.ofLEByteArray data
+    let x <- f32 f
+    return x.toLEByteArray
+  | .float64 => do
+    let f <- Float.ofLEByteArray data
+    let x <- f64 f
+    return x.toLEByteArray
+  | _ => throw "operation requires a float type"
+
+-- I'm sure this exists but couldn't find it
```
govereau, line 479: It is `Function.comp`, not easy to find; I expected it to not be hidden in a namespace.
There is also a unicode infix operator, which is `\circ` for me (when I remember).

TensorLib/Dtype.lean
```diff
@@ -437,6 +451,131 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err
...
+type before calling the function. This follows some rules, which we approximate here, given we don't
+have all the types available in NumPy (e.g. float16) and we may have types like bfloat that aren't
+in NumPy out of the box.
+-/
+def floatVariant (dtype : Dtype) : Dtype :=
+  if dtype.itemsize <= 1 then Dtype.float32 -- leaving this branch in because should be Dtype.float16. TODO to change
+  else if dtype.itemsize <= 2 then Dtype.float32
+  else Dtype.float64
+
+private def liftFloatUnop (f32 : Float32 -> Err Float32) (f64 : Float -> Err Float)
```
govereau, line 465: How many of these functions will we have long term? Is it just the two?

govereau (approved):

TensorLib/Dtype.lean
seanmcl: Thanks.

TensorLib/Dtype.lean
seanmcl: One for each floating point type. We need (badly) float16 but there are probably 3 others. We can have a struct at some point. I can do that now if you like? I thought about having a class but that seems messy; I like how we've created an essentially untyped language of ByteArray, compared with the heavy class-based system we had before.

govereau: I think this way is better as well, just wondering how big the list would get. Structure is a good idea, but probably not necessary until `count > 2`.

PR #62, opened 2025-04-01
feat: add shift ops

seanmcl:


TensorLib/Common.lean
```diff
@@ -117,6 +117,12 @@ def _root_.ByteArray.toInt (arr : ByteArray) : Int := Id.run do
 
 def _root_.ByteArray.toBool (arr : ByteArray) : Bool := arr.data.any fun v => v != 0
 
+def _root_.ByteArray.take (arr : ByteArray) (n : Nat) : ByteArray :=
+  ByteArray.mk (arr.data.take n)
+
+def _root_.ByteArray.drop (arr : ByteArray) (n : Nat) : ByteArray :=
+  ByteArray.mk (arr.data.drop n)
```
govereau, line 124: Minor comment, but I wonder if `ByteArray.extract` would be more efficient here?

govereau (approved):

TensorLib/Common.lean
seanmcl: Drop is implemented as a one liner to extract

PR #66, opened 2025-04-03
feat: add `np.full`

seanmcl:
Also simplifies and fixes a bug with `np.ones` that gave the wrong bits for floating point.

### DIFF
--- a/TensorLib/Tensor.lean
+++ b/TensorLib/Tensor.lean
@@ -99,15 +99,6 @@ def zeros (dtype : Dtype) (shape : Shape) : Tensor := Id.run do
     data := data.push 0
   { dtype := dtype, shape := shape, data := data }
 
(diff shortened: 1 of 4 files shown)


TensorLib/Tensor.lean
```diff
@@ -99,15 +99,6 @@ def zeros (dtype : Dtype) (shape : Shape) : Tensor := Id.run do
     data := data.push 0
   { dtype := dtype, shape := shape, data := data }
 
-def ones (dtype : Dtype) (shape : Shape) : Tensor := Id.run do
-  let size := dtype.itemsize * shape.count
-  let itemsize := dtype.itemsize
-  let mut data := ByteArray.mkEmpty size
-  for i in [0:size] do
-    let byte := if i.mod itemsize == 0 then 1 else 0
```
seanmcl, line 107: This was the bug; floats don't have the same representation of 1.

govereau (approved):

PR #69, opened 2025-05-01
chore: Refactor Common.lean, II

seanmcl:
Another refactor, just enough to pull out the Float code to its own module.


TensorLib/Float.lean
```diff
@@ -0,0 +1,84 @@
...
+def Float32.ofLEByteArray! (arr : ByteArray) : Float32 := get! $ Float32.ofLEByteArray arr
+
+def Float.ofLEByteArray (arr : ByteArray) : Err Float := do
+  return Float.ofBits (<- arr.toUInt64LE)
+
+def Float.ofLEByteArray! (arr : ByteArray) : Float := get! $ Float.ofLEByteArray arr
+
+def _root_.Float32.toNat (f : Float32) : Nat := f.toUInt64.toNat
+
+def _root_.Float32.toInt (f : Float32) : Int :=
```
govereau, line 48: I checked and it looks like this returns 0 for NaN. Seems fine, but maybe worth a comment?

govereau (approved):

TensorLib/Float.lean
seanmcl: Good point. Added a #guard as documentation

PR #72, opened 2025-05-01
feat: Add big-endian versions of util functions for UIntX and ByteArray

seanmcl:
Needed by CBOR in KLR, for example.


govereau (approved):
Awesome, thanks.

PR #74, opened 2025-05-05
Fix intAtDimIndex to return the right value

aqjune-aws:
This patch fixes intAtDimIndex to return the correct value at the location, by multiplying the size of element in dimIndexToPosition.

This makes the new `#guard` statements pass in this pull request.


seanmcl (approved):
Ah, good find. Very annoying to need to worry about the item size in all these indexing operations. Maybe there's a good more general solution someday.

aqjune-aws:
Hi - I don't have permission to merge this. Could someone have this instead for me please?

PR #75, opened 2025-05-07
chore: fix bug in array utils and add a monadlift instance

seanmcl:


TensorLib/Common.lean
```diff
@@ -12,7 +12,12 @@ export Function(comp)
 
 abbrev Err := Except String
 
-def impossible {a : Type} [h : Inhabited a] (msg : String := "") := @panic a h s!"Invariant violation: {msg}"
+instance : MonadLiftT Option Err where
```
govereau, line 15: I think the manual recommends `MonadLift` for instances.
https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#MonadLift

govereau (approved):

PR #91, opened 2025-10-08
fix: Resolving name clashing issue with size_push theorem and syntax error

segevem:
When using TensorLib and a dependency which depends on Batteries, the `size_push` theorem in Batteries clashes with the one defined in TensorLib. Fixed by removing the one in TensorLib and importing Batteries' version of `ByteArray` to supply the `size_push` theorem.

Removed a space behind a match expression in Iterator.lean which caused a syntax error.

### DIFF
--- a/TensorLib/ByteArray.lean
+++ b/TensorLib/ByteArray.lean
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -/
 
+import Batteries.Data.ByteArray
 import Plausible
 import TensorLib.Bytes
 import TensorLib.Common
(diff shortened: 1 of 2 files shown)


TensorLib/ByteArray.lean
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 import Plausible
 import TensorLib.Bytes
 import TensorLib.Common
+import Batteries.Data.ByteArray
```
seanmcl: nit: please sort this (ie put it above Plausible)

TensorLib/ByteArray.lean
```diff
@@ -169,9 +170,6 @@ def _root_.ByteArray.replicate (n : Nat) (v : UInt8) : ByteArray := Id.run do
 
 #guard ByteArray.replicate 5 7 == ⟨ #[7, 7, 7, 7, 7] ⟩
 
-@[simp] theorem _root_.ByteArray.size_push (a : ByteArray) (b : UInt8) : (a.push b).size = a.size + 1 :=
```
seanmcl, line 172: so just removing this theorem does what again?

seanmcl: Ah I see, putting the theorem in the root namespace messed you up. Got it!

seanmcl (approved):

seanmcl:
Thanks again. Please squash your commits and we can merge.

segevem:
Thank you! Squashed.

PR #92, opened 2026-06-23
Fp16 Dtype

SmoothThunk:
Added support for fp16 in TensorLib.

### DIFF
--- a/TensorLib/Dtype.lean
+++ b/TensorLib/Dtype.lean
@@ -38,6 +38,7 @@ inductive Dtype where
 | uint64
 | float32
 | float64
+| float16
 deriving BEq, Repr, Inhabited
 
 
@@ -79,6 +80,7 @@ instance : ToString Dtype where
   | uint64 => "uint64"
   | float32 => "float32"
   | float64 => "float64"
+  | float16 => "float16"
 
 def isOneByte (x : Dtype) : Bool := match x with
 | bool | int8 | uint8 => true
@@ -104,7 +106,7 @@ def isFloat (x : Dtype) : Bool := match x with
 def itemsize (x : Dtype) : Nat := match x with
 | float64 | int64 | uint64 => 8
 | float32 | int32 | uint32 => 4
-| int16 | uint16 => 2
+| float16 | int16 | uint16 => 2
 | bool | int8 | uint8 => 1
 
 -- This is the type NumPy returns when using binary operators on arrays
@@ -113,6 +115,11 @@ def join (x y : Dtype) : Option Dtype :=
   let (x, y) := if x.itemsize <= y.itemsize then (x, y) else (y, x)
   if x == y then x else
   match x, y with
+  | float16, float16 => float16
+  | float16, float32 => float32
+  | float16, float64 => float64
+  | float16, _ => none
+  | _, float16 => none
   | float32, float64 => float64
   | float32, _
   | _, float32
@@ -135,6 +142,7 @@ def join (x y : Dtype) : Option Dtype :=
   | _, int64
   | uint64, _ => none -- NumPy gives "can't safely coerce" for uint64/int64
 
+-- Can we cast from one dtype to another without losing information
 def lossless (fromDtype toDtype : Dtype) : Bool := match fromDtype, toDtype with
 | .bool, _ => true
 | _, .bool => false
@@ -181,6 +189,10 @@ def lossless (fromDtype toDtype : Dtype) : Bool := match fromDtype, toDtype with
 | .int64, _ => false
 | .uint64, .uint64 => true
 | .uint64, _ => false
+| .float16, .float16
+| .float16, .float32
+| .float16, .float64 => true
+| .float16, _ => false
 | .float32, .float32
 | .float32, .float64 => true
 | .float32, _ => false
@@ -224,6 +236,7 @@ private def maxSafeNat : Dtype -> Option Nat
 | .int32 => some 0x7FFFFFFF
 | .uint64 => some 0xFFFFFFFFFFFFFFFF
 | .int64 => some 0x7FFFFFFFFFFFFFFF
+| .float16 => maxSafeNatForFloat16
 | .float32 => maxSafeNatForFloat32
 | .float64 => maxSafeNatForFloat64
 
@@ -241,6 +254,7 @@ private def minSafeInt : Dtype -> Option Int
 | .int16 => some (-0x8000)
 | .int32 => some (-0x80000000)
 | .int64 => some (-0x8000000000000000)
+| .float16 => some (-maxSafeNatForFloat16)
 | .float32 => some (-maxSafeNatForFloat32)
 | .float64 => some (-maxSafeNatForFloat64)
 
@@ -260,6 +274,7 @@ def byteArrayOfNatOverflow (dtype : Dtype) (n : Nat) : ByteArray := match dtype 
 | .int32 => toLEByteArray n.toInt32
 | .uint64 => toLEByteArray n.toUInt64
 | .int64 => toLEByteArray n.toInt64
+| .float16 => toLEByteArray n.toFloat32.toFloat16Bits
 | .float32 => toLEByteArray n.toFloat32
 | .float64 => toLEByteArray n.toFloat
 
@@ -287,6 +302,7 @@ private def byteArrayOfIntOverflow (dtype : Dtype) (n : Int) : ByteArray := matc
 | .uint16 | .int16 => toLEByteArray n.toInt16
 | .uint32 | .int32 => toLEByteArray n.toInt32
 | .uint64 | .int64 => toLEByteArray n.toInt64
+| .float16 => toLEByteArray n.toFloat32.toFloat16Bits
 | .float32 => toLEByteArray n.toFloat32
 | .float64 => toLEByteArray n.toFloat64
 
@@ -362,6 +378,13 @@ private def byteArrayToFloat32RoundTrip (dtype : Dtype) (f : Float32) : Bool :=
 #guard float32.byteArrayToFloat32RoundTrip (Float32.sqrt 2)
 #guard float32.byteArrayToFloat32RoundTrip 0
 
+
+-- Read 2 bytes as fp16, convert to fp32
+-- perform arithmetic computations since lean doesnt have native fp16 and finally downcast to fp16 to store result
+def byteArrayToFloat16AsFloat32 (arr : ByteArray) : Err Float32 := do
+  let bits ← arr.toUInt16LE -- Read 2 bytes as UInt16 in little Endian
+  return bits.toFloat32FromFloat16
+
 /-
 NumPy addition overflows and underflows without complaint. We will do the same.
 -/
@@ -378,6 +401,10 @@ def add (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
     return dtype.byteArrayOfNatOverflow (x.toNat + y.toNat)
   | .int8 | .int16| .int32 | .int64 => do
     dtype.byteArrayOfInt (x.toInt + y.toInt)
+  | .float16 => do -- read inputs as fp32
+    let x <- byteArrayToFloat16AsFloat32 x
+    let y <- byteArrayToFloat16AsFloat32 y
+    return toLEByteArray (x + y).toFloat16Bits -- add in fp32 then downcast to fp16
   | .float32 => do
     let x <- dtype.byteArrayToFloat32 x
     let y <- dtype.byteArrayToFloat32 y
@@ -400,6 +427,10 @@ def sub (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
     let x <- dtype.byteArrayToFloat32 x
     let y <- dtype.byteArrayToFloat32 y
     dtype.byteArrayOfFloat32 (x - y)
+  | .float16 => do
+    let x <- byteArrayToFloat16AsFloat32 x
+    let y <- byteArrayToFloat16AsFloat32 y
+    return toLEByteArray (x - y).toFloat16Bits -- sub in fp32 then store result in fp16
   | .float64 => do
     let x <- dtype.byteArrayToFloat64 x
     let y <- dtype.byteArrayToFloat64 y
@@ -419,6 +450,10 @@ def mul (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
     let x <- dtype.byteArrayToFloat32 x
     let y <- dtype.byteArrayToFloat32 y
     dtype.byteArrayOfFloat32 (x * y)
+  | .float16 => do
+    let x <- byteArrayToFloat16AsFloat32 x
+    let y <- byteArrayToFloat16AsFloat32 y
+    return toLEByteArray (x * y).toFloat16Bits -- multiply in fp32 then downcast to fp16 to store
   | .float64 => do
     let x <- dtype.byteArrayToFloat64 x
     let y <- dtype.byteArrayToFloat64 y
@@ -438,6 +473,10 @@ def div (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
     let x <- dtype.byteArrayToFloat32 x
     let y <- dtype.byteArrayToFloat32 y
     dtype.byteArrayOfFloat32 (x / y)
+  | .float16 => do
+    let x <- byteArrayToFloat16AsFloat32 x
+    let y <- byteArrayToFloat16AsFloat32 y
+    return toLEByteArray (x / y).toFloat16Bits -- divide in fp32 then downcast to fp16
   | .float64 => do
     let x <- dtype.byteArrayToFloat64 x
     let y <- dtype.byteArrayToFloat64 y
@@ -455,6 +494,9 @@ def abs (dtype : Dtype) (x : ByteArray) : Err ByteArray := do
   match dtype with
   | .uint8 | .uint16 | .uint32 | .uint64 => return x
   | .int8 | .int16| .int32 | .int64 => return dtype.byteArrayOfIntOverflow x.toInt.natAbs
+  | .float16 => do
+    let x <- byteArrayToFloat16AsFloat32 x
+    return toLEByteArray x.abs.toFloat16Bits
   | .float32 =>
     let x <- dtype.byteArrayToFloat32 x
     dtype.byteArrayOfFloat32 x.abs
@@ -488,6 +530,29 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err 
   | .float32, .float64 => do
     let f <- Float32.ofLEByteArray data
     return toLEByteArray f.toFloat
+   -- float16 to integer types: change to float32, then convert to int/nat
+  | .float16, .uint8 | .float16, .uint16 | .float16, .uint32 | .float16, .uint64 => do
+    let f ← byteArrayToFloat16AsFloat32 data
+    return toDtype.byteArrayOfNatOverflow f.toNat
+  | .float16, .int8 | .float16, .int16 | .float16, .int32 | .float16, .int64 => do
+    let f ← byteArrayToFloat16AsFloat32 data
+    return toDtype.byteArrayOfIntOverflow f.toInt
+  -- float16 to float32/float64: decode to float32, then widen if needed
+  | .float16, .float32 => do
+    let f ← byteArrayToFloat16AsFloat32 data
+    return toLEByteArray f
+  | .float16, .float64 => do
+    let f ← byteArrayToFloat16AsFloat32 data
+    return toLEByteArray f.toFloat
+  -- float32/float64 → float16: decode, then downcast to float16 bits
+  | .float32, .float16 => do
+    let f ← Float32.ofLEByteArray data
+    return toLEByteArray f.toFloat16Bits
+  | .float64, .float16 => do
+    let f ← Float.ofLEByteArray data
+    return toLEByteArray f.toFloat32.toFloat16Bits
+  -- float16 to float16: already handled by the `if fromDtype == toDtype` check above
+  | .float16, .float16 => impossible
   | .float64, .float32 => do
     let f <- Float.ofLEByteArray data
     return toLEByteArray f.toFloat32
@@ -504,6 +569,9 @@ def isZero (dtype : Dtype) (x : ByteArray) : Err Bool := match dtype with
 | int64
 | uint64 => return x.data.all fun v => v == 0
 -- We need to worry about -0, which is not all 0s in the bit pattern.
+| float16 => do
+  let f <- byteArrayToFloat16AsFloat32 x
+  return f == 0
 | float32 => do
   let f <- Float32.ofLEByteArray x
   return f == 0
@@ -626,7 +694,7 @@ def tanh : Dtype -> ByteArray -> Err ByteArray :=
 def tanh! (dtype : Dtype) (data : ByteArray) : ByteArray := get! $ tanh dtype data
 
 private def shift (f : UInt64 -> UInt64 -> UInt64) (dtype : Dtype) (bits : ByteArray) (shiftAmount : ByteArray) : Err ByteArray := match dtype with
-| .float32 | .float64 => throw "shifts not supported at float type"
+| .float32 | .float64 | .float16 => throw "shifts not supported at float type"
 | .bool => throw "In NumPy, bool shifts are cast to int64. This seems arbitrary so please cast (e.g. with astype) before you shift."
 | .uint64 | .int64 | .uint32 | .int32 | .uint16 | .int16 | .uint8 | .int8 =>
   let k := dtype.itemsize
--- a/TensorLib/Dtype.lean
+++ b/TensorLib/Dtype.lean
@@ -117,7 +117,6 @@ def join (x y : Dtype) : Option Dtype :=
   let (x, y) := if x.itemsize <= y.itemsize then (x, y) else (y, x)
   if x == y then x else
   match x, y with
-  | float16, float16 => float16
   | float16, float32 => float32
   | float16, float64 => float64
   | float16, int16 => float32
@@ -126,12 +125,8 @@ def join (x y : Dtype) : Option Dtype :=
   | float16, uint32 => float64
   | float16, int64 => float64
   | float16, uint64 => float64
-  | float16, bool => float16
-  | int8, float16 => float16
-  | float16, int8 => float16
-  | uint8, float16 => float16
-  | float16, uint8 => float16
   | float32, float64 => float64
+  | float16, _ => float16 -- for all other cases that promote to fp16
   | float32, _
   | _, float32
   | float64, _
@@ -361,8 +356,6 @@ def byteArrayOfFloat16 (dtype : Dtype) (f : Float32) : Err ByteArray :=
   if dtype == .float16 then .ok (toLEByteArray f.toFloat16Bits)
   else .error "Illegal type conversion"
 
-private def byteArrayOfFloat16! (dtype : Dtype) (f : Float32) : ByteArray := get! $ byteArrayOfFloat16 dtype f
-
 private def byteArrayToFloat64RoundTrip (dtype : Dtype) (f : Float) : Bool :=
   let res := do
     let arr <- dtype.byteArrayOfFloat64 f
@@ -402,7 +395,6 @@ def byteArrayToFloat16 (dtype : Dtype)(arr : ByteArray) : Err Float32 := match d
   | .float16 => arr.toUInt16LE.map UInt16.toFloat32FromFloat16
   | _ => .error "Illegal type conversion"
 
-def byteArrayToFloat16! (dtype : Dtype) (arr : ByteArray) : Float32 := get! $ byteArrayToFloat16 dtype arr
 
 
 private def byteArrayToFloat16RoundTrip (dtype : Dtype) (f : Float32) : Bool :=
@@ -418,12 +410,6 @@ private def byteArrayToFloat16RoundTrip (dtype : Dtype) (f : Float32) : Bool :=
 #guard float16.byteArrayToFloat16RoundTrip 42
 #guard float16.byteArrayToFloat16RoundTrip (-0)
 
-
--- Read 2 bytes as fp16, convert to fp32
--- perform arithmetic computations since lean doesnt have native fp16 and finally downcast to fp16 to store result
-def byteArrayToFloat16AsFloat32 (arr : ByteArray) : Err Float32 := do
-  let bits <- arr.toUInt16LE -- Read 2 bytes as UInt16 in little Endian
-  return bits.toFloat32FromFloat16
 
 -- Add produces the IEEE754 result because fp32 (p=24) satisfies the innocuous double rounding condition (theoreom 20)
 -- https://hal.science/hal-01091186v1/document
@@ -562,7 +548,7 @@ def isZero (dtype : Dtype) (x : ByteArray) : Err Bool := match dtype with
 | uint64 => return x.data.all fun v => v == 0
 -- We need to worry about -0, which is not all 0s in the bit pattern.
 | float16 => do
-  let f <- byteArrayToFloat16AsFloat32 x
+  let f <- Dtype.byteArrayToFloat16 .float16 x
   return f == 0
 | float32 => do
   let f <- Float32.ofLEByteArray x
@@ -605,17 +591,17 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err 
     return toLEByteArray f.toFloat
    -- float16 to integer types: change to float32, then convert to int/nat
   | .float16, .uint8 | .float16, .uint16 | .float16, .uint32 | .float16, .uint64 => do
-    let f <- byteArrayToFloat16AsFloat32 data
+    let f <- Dtype.byteArrayToFloat16 .float16 data
     return toDtype.byteArrayOfNatOverflow f.toNat
   | .float16, .int8 | .float16, .int16 | .float16, .int32 | .float16, .int64 => do
-    let f <- byteArrayToFloat16AsFloat32 data
+    let f <- Dtype.byteArrayToFloat16 .float16 data
     return toDtype.byteArrayOfIntOverflow f.toInt
   -- float16 to float32/float64: decode to float32, then widen if needed
   | .float16, .float32 => do
-    let f <- byteArrayToFloat16AsFloat32 data
+    let f <- Dtype.byteArrayToFloat16 .float16 data
     return toLEByteArray f
   | .float16, .float64 => do
-    let f <- byteArrayToFloat16AsFloat32 data
+    let f <- Dtype.byteArrayToFloat16 .float16 data
     return toLEByteArray f.toFloat
   -- float32/float64 → float16: decode, then downcast to float16 bits
   | .float32, .float16 => do
@@ -988,8 +974,10 @@ warning: declaration uses 'sorry'
 #guard_msgs in
 -- Lossless translations should be OK
 example (fromDtype toDtype : Dtype) (n : Nat) :
+-- only try values within fromDtye's representable range
+-- fp16 max n = 2^11, fp32 max n = 2^24, fp64 = 2^53, uint8 = 255, int8 = 127, etc
   if lossless fromDtype toDtype then
-    canCastLosslessIntRoundTrip fromDtype n toDtype
+    if n > fromDtype.maxSafeNat.getD n then true else canCastLosslessIntRoundTrip fromDtype n toDtype
   else true := by plausible
 
 
@@ -1018,7 +1006,8 @@ example (a : UInt16) :
   let zero := toLEByteArray (0 : UInt16)
   let f := a.toFloat32FromFloat16
    -- f != f is true only for NaN; check identity for everything else
-  f != f ∨ Dtype.add .float16 xa zero == .ok xa := by plausible
+   -- if value is +-0 skip byte equality check since addition is identity upto sign
+  f != f ∨ f == 0 ∨ Dtype.add .float16 xa zero == .ok xa := by plausible
 
 
 #guard Dtype.uint8.leftShift! (ByteArray.mk #[10]) (ByteArray.mk #[1]) == ByteArray.mk #[20]
--- a/TensorLib/Dtype.lean
+++ b/TensorLib/Dtype.lean
@@ -39,7 +39,7 @@ inductive Dtype where
 | float16
 | float32
 | float64
-deriving BEq, Repr, Inhabited
+deriving BEq, Repr, Inhabited, DecidableEq
 
 
 namespace Dtype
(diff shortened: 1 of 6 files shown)


seanmcl:
Please make sure you squash all these commits when you merge.

TensorLib/Dtype.lean
```diff
@@ -38,6 +38,7 @@ inductive Dtype where
 | uint64
 | float32
 | float64
+| float16
```
seanmcl: move above float32

TensorLib/Dtype.lean
```diff
@@ -362,6 +378,13 @@ private def byteArrayToFloat32RoundTrip (dtype : Dtype) (f : Float32) : Bool :=
 #guard float32.byteArrayToFloat32RoundTrip (Float32.sqrt 2)
 #guard float32.byteArrayToFloat32RoundTrip 0
 
+
+-- Read 2 bytes as fp16, convert to fp32
+-- perform arithmetic computations since lean doesnt have native fp16 and finally downcast to fp16 to store result
```
seanmcl: Are we sure this is the correct math for fp16?

TensorLib/Dtype.lean
```diff
@@ -488,6 +530,29 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err
...
+    let f ← byteArrayToFloat16AsFloat32 data
+    return toLEByteArray f.toFloat
+  -- float32/float64 → float16: decode, then downcast to float16 bits
+  | .float32, .float16 => do
+    let f ← Float32.ofLEByteArray data
+    return toLEByteArray f.toFloat16Bits
+  | .float64, .float16 => do
+    let f ← Float.ofLEByteArray data
+    return toLEByteArray f.toFloat32.toFloat16Bits
+  -- float16 to float16: already handled by the `if fromDtype == toDtype` check above
```
seanmcl: move this to the bottom match

govereau (approved):
Looks good to me.
May be worth checking if the conversion to and from f16 matches what S0 semantics will want.

TensorLib/Dtype.lean
SmoothThunk: Moved fp16 above fp32.

TensorLib/Dtype.lean
SmoothThunk: Done

TensorLib/Dtype.lean
SmoothThunk: For now I have implemented it as fp16 inputs are upcast to fp32, add in fp32 and then downcast to fp16 by rebiasing exponent + right shifting mantissa by 13 so the precision I get in the result (fp16) is rounded to the nearest value that can be represented in 10 bit mantissa.

TensorLib/Float.lean
```diff
@@ -96,6 +99,44 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+-- left shift mantissa 13
+-- pack it into 32 bit layout
+def _root_.UInt16.toFloat32FromFloat16 (bits : UInt16) : Float32 :=
+  let sign := (bits >>> 15) &&& 1  -- sign bit; positive = 0, negative = 1
+  let exp := (bits >>> 10) &&& 0x1F -- extract bits 14..10
+  let mant := bits &&& 0x3FF -- mantissa bits
+  if exp ==  0 && mant == 0 then
+    if sign == 1 then Float32.ofBits 0x80000000 -- -0
+    else Float32.ofBits 0  -- +0
+  else if exp == 0x1FF then -- exp is all 1's so this is either +-inf or NaN
```
seanmcl: **Bug (high): the Inf/NaN decode branch is dead code.** `exp` is masked to 5 bits at line 126 (`&&& 0x1F`, max `0x1F`), so `exp == 0x1FF` is never true. fp16 `+Inf` (`0x7C00`) falls through to the normal `else` branch -> `newExp = 0x1F - 15 + 127 = 143` -> `Float32.ofBits 0x47800000 = 65536.0` instead of `+Inf`; NaN `0x7E00` decodes to ~98304.0. Every fp16 op that consumes an Inf/NaN silently yields a wrong finite value. Should be `exp == 0x1F`.

TensorLib/Float.lean
```diff
@@ -96,6 +99,44 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+  let sign := (bits >>> 31) &&& 1
+  let exp := (bits >>> 23) &&& 0xFF
+  let mant := bits &&& 0x7FFFFF
+  let sign16 := sign.toUInt16 <<< 15
+  if exp == 0 then sign16 -- 0 or subnormal float32 is 0 in float16
+  else if exp == 0xFF then sign16 ||| ((0x1F : UInt16) <<< 10) ||| (mant >>> 13).toUInt16 -- if Inf or NaN. 0x1F
+  else
+    let newExp : Int := exp.toNat - 127 + 15 -- rebias from float32 to float16 bias
+    if newExp >= 0x1F then sign16 ||| ((0x1F : UInt16) <<< 10) -- overflow -> inf
+    else if newExp <= 0 then sign16 -- underflow -> 0
```
seanmcl: **Bug (high): the encoder flushes the entire fp16 subnormal range to zero.** `else if newExp <= 0 then sign16` discards values that are representable as fp16 subnormals (down to 2^-24). e.g. fp32 `2^-15` (`0x38000000`), representable as fp16 `0x0200`, has `newExp = 112 - 127 + 15 = 0 <= 0` -> returns just `sign16` -> value silently becomes `0.0`.

TensorLib/Float.lean
```diff
@@ -96,6 +99,44 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+  if exp ==  0 && mant == 0 then
+    if sign == 1 then Float32.ofBits 0x80000000 -- -0
+    else Float32.ofBits 0  -- +0
+  else if exp == 0x1FF then -- exp is all 1's so this is either +-inf or NaN
+    if mant == 0 then Float32.ofBits (sign.toUInt32 <<< 32 ||| 0x7F800000) -- +- inf
+    else Float32.ofBits 0x7FC00000 -- Nan
+  else
+    -- Rebias the exponent from float16 (15) bias to float32 (127)
+    -- shift mant from 10 bits to 23 bits so pad with 13 0's
+    let newExp := exp.toUInt32 - 15 + 127
```
seanmcl, line 185: **Bug (high): the decoder has no subnormal path.** fp16 `exp==0, mant!=0` is not special-cased (only `exp==0 && mant==0` returns zero), so it reaches here, where `exp.toUInt32 - 15` underflows (unsigned wrap) and there is no denormal scaling. fp16 `0x0001` (true value 2^-24 ~= 5.96e-8) decodes to a garbage magnitude rather than the correct tiny value.

TensorLib/Float.lean
```diff
@@ -96,6 +99,44 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+  let exp := (bits >>> 23) &&& 0xFF
+  let mant := bits &&& 0x7FFFFF
+  let sign16 := sign.toUInt16 <<< 15
+  if exp == 0 then sign16 -- 0 or subnormal float32 is 0 in float16
+  else if exp == 0xFF then sign16 ||| ((0x1F : UInt16) <<< 10) ||| (mant >>> 13).toUInt16 -- if Inf or NaN. 0x1F
+  else
+    let newExp : Int := exp.toNat - 127 + 15 -- rebias from float32 to float16 bias
+    if newExp >= 0x1F then sign16 ||| ((0x1F : UInt16) <<< 10) -- overflow -> inf
+    else if newExp <= 0 then sign16 -- underflow -> 0
+    else sign16 ||| (newExp.toNat.toUInt16 <<< 10) ||| (mant >>> 13).toUInt16
```
seanmcl: **Bug (medium): truncation instead of round-to-nearest-even.** `mant >>> 13` truncates the discarded mantissa bits; NumPy uses round-half-to-even for fp32->fp16. Any value whose discarded low 13 bits are at/above the rounding threshold ends up one ULP low, so round-trip and arithmetic comparisons against NumPy will mismatch.

TensorLib/Float.lean
```diff
@@ -96,6 +99,44 @@ def _root_.Float.toInt (f : Float) : Int :=
...
 
+-- Convert a float32 to Float16 as UInt16
+def _root_.Float32.toFloat16Bits (f : Float32) : UInt16 :=
+  let bits := f.toBits
+  let sign := (bits >>> 31) &&& 1
+  let exp := (bits >>> 23) &&& 0xFF
+  let mant := bits &&& 0x7FFFFF
+  let sign16 := sign.toUInt16 <<< 15
+  if exp == 0 then sign16 -- 0 or subnormal float32 is 0 in float16
+  else if exp == 0xFF then sign16 ||| ((0x1F : UInt16) <<< 10) ||| (mant >>> 13).toUInt16 -- if Inf or NaN. 0x1F
```
seanmcl: **Bug (medium): a low-payload NaN can collapse to Inf.** The Inf/NaN encode path sets the fp16 mantissa to `(mant >>> 13)`, which is `0` when the fp32 NaN payload lives only in the low 13 bits. fp32 signaling NaN `0x7F800001` -> fp16 `0x7C00` (`+Inf`), losing NaN-ness.

TensorLib/Float.lean
```diff
@@ -96,6 +99,44 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+-- pack it into 32 bit layout
+def _root_.UInt16.toFloat32FromFloat16 (bits : UInt16) : Float32 :=
+  let sign := (bits >>> 15) &&& 1  -- sign bit; positive = 0, negative = 1
+  let exp := (bits >>> 10) &&& 0x1F -- extract bits 14..10
+  let mant := bits &&& 0x3FF -- mantissa bits
+  if exp ==  0 && mant == 0 then
+    if sign == 1 then Float32.ofBits 0x80000000 -- -0
+    else Float32.ofBits 0  -- +0
+  else if exp == 0x1FF then -- exp is all 1's so this is either +-inf or NaN
+    if mant == 0 then Float32.ofBits (sign.toUInt32 <<< 32 ||| 0x7F800000) -- +- inf
```
seanmcl: **Bug: wrong shift width.** `sign.toUInt32 <<< 32` — shifting a UInt32 by 32 wraps the shift amount to 0 in Lean, so the sign lands in bit 0, not bit 31. Should be `<<< 31` (as used on line 138). Latent only because the enclosing branch is unreachable today (the `0x1FF` test above); fixing that test exposes this.

TensorLib/Dtype.lean
```diff
@@ -626,7 +694,7 @@ def tanh : Dtype -> ByteArray -> Err ByteArray :=
 def tanh! (dtype : Dtype) (data : ByteArray) : ByteArray := get! $ tanh dtype data
 
 private def shift (f : UInt64 -> UInt64 -> UInt64) (dtype : Dtype) (bits : ByteArray) (shiftAmount : ByteArray) : Err ByteArray := match dtype with
-| .float32 | .float64 => throw "shifts not supported at float type"
+| .float32 | .float64 | .float16 => throw "shifts not supported at float type"
```
seanmcl, line 745: **Bug (medium) — re: `isFloat` (line 101), anchored here.** You correctly added `.float16` to this shift guard, but `isFloat` (line 101) was *not* updated and still returns true only for float32/float64. `bitwiseBinop` (line 725) and `bitwiseUnop` (761) guard with `if xtyp.isFloat then throw`; since `isFloat .float16 == false`, bitwise And/Or/Xor/Not run on raw fp16 bytes — which NumPy rejects ("ufunc not supported for float16"). Add `.float16` to `isFloat`.

TensorLib/Dtype.lean
```diff
@@ -113,6 +115,11 @@ def join (x y : Dtype) : Option Dtype :=
   let (x, y) := if x.itemsize <= y.itemsize then (x, y) else (y, x)
   if x == y then x else
   match x, y with
+  | float16, float16 => float16
+  | float16, float32 => float32
+  | float16, float64 => float64
+  | float16, _ => none
```
seanmcl: **Bug (medium): `join` rejects float16-vs-integer.** `| float16, _ => none` / `| _, float16 => none` blanket-reject every fp16-vs-integer pair, contradicting join's documented "type NumPy returns for binary operators" contract. NumPy: `result_type(float16, int8) = float16`, `(float16, int16) = float32`, `(float16, int32/int64) = float64`. All return `none` here.

TensorLib/Dtype.lean
```diff
@@ -181,6 +189,10 @@ def lossless (fromDtype toDtype : Dtype) : Bool := match fromDtype, toDtype with
 | .int64, _ => false
 | .uint64, .uint64 => true
 | .uint64, _ => false
+| .float16, .float16
```
seanmcl, line 204: **Bug (medium) — re: casts *into* float16.** These arms only cover the FROM-float16 direction. `int8 -> float16` falls through to `| .int8, _ => false` (line 155) and `uint8 -> float16` to `| .uint8, _ => false` (line 165), but every int8/uint8 value is exactly representable in fp16 (NumPy `can_cast(..., 'safe')` is True). (int16/uint16 -> float16 correctly stay false.)

TensorLib/Dtype.lean
seanmcl: **Coverage gap — re: `gen` (line 51).** `.float16` was not added to `gen : Gen Dtype` (the list ends at float64), so all Plausible / `SampleableExt` property checks silently skip float16 and the new paths get no generative coverage. Adding it here would mean the round-trip / join / lossless property checks exercise fp16.

TensorLib/Dtype.lean
```diff
@@ -378,6 +401,10 @@ def add (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
     return dtype.byteArrayOfNatOverflow (x.toNat + y.toNat)
   | .int8 | .int16| .int32 | .int64 => do
     dtype.byteArrayOfInt (x.toInt + y.toInt)
+  | .float16 => do -- read inputs as fp32
```
seanmcl, line 436: **Cleanup: missing symmetric helpers.** The float32/float64 paths route through `dtype.byteArrayToFloat32` / `byteArrayOfFloat32` (with size+dtype validation and a round-trip `#guard`). The five fp16 arms (add 407, sub 433, mul 456, div 479, abs 499) hand-roll `byteArrayToFloat16AsFloat32` + `toLEByteArray (...).toFloat16Bits` inline. A `byteArrayToFloat16` / `byteArrayOfFloat16` pair plus a `byteArrayToFloat16RoundTrip` `#guard` would give one source of truth instead of five edit sites — and the round-trip guard would have caught the conversion bugs above.

TensorLib/Dtype.lean
```diff
@@ -488,6 +530,29 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err
   | .float32, .float64 => do
     let f <- Float32.ofLEByteArray data
     return toLEByteArray f.toFloat
+   -- float16 to integer types: change to float32, then convert to int/nat
+  | .float16, .uint8 | .float16, .uint16 | .float16, .uint32 | .float16, .uint64 => do
```
seanmcl, line 597: **Bug (low, pre-existing but newly reachable for fp16): `-0.0` casts to `True`.** The `_, .bool` arm (line 513) tests raw bytes `all == 0`; fp16 `-0.0` is `00 80` (nonzero sign byte) -> returns `1`. NumPy: `np.float16(-0.0).astype(bool)` is `False`. Note `isZero` got an explicit `-0` fix but this cast path did not (also affects float32/float64).

TensorLib/Npy.lean
```diff
@@ -108,6 +109,7 @@ def dtypeNameToNpyString (t : TensorLib.Dtype) : String := match t with
 | .uint16 => "u2"
 | .uint32 => "u3"
 | .uint64 => "u4"
+| .float16 => "f2"
```
seanmcl, line 112: The new `float16 -> "f2"` is correct. **Pre-existing bug in this function (flagging since the PR edits it):** `uint32 -> "u3"` (line 110) and `uint64 -> "u4"` (line 111) are wrong — should be `"u4"` and `"u8"`. Neither round-trips with `dtypeNameFromNpyString` (which has no `"u3"` case, and maps `"u4"` back to uint32), so writing then reading a uint32/uint64 .npy header fails to parse or mis-reads the width. Unrelated to fp16.

TensorLib/Test.lean
```diff
@@ -39,7 +39,8 @@ private def testTensorElementBV (dtype : Dtype) : IO Bool := do
 
 def runAllTests : IO Bool := do
   return (<- testTensorElementBV Dtype.uint16) &&
-         (<- testTensorElementBV Dtype.uint32)
+         (<- testTensorElementBV Dtype.uint32) &&
+         (<- testTensorElementBV Dtype.float16)
```
seanmcl: **Test gap: this case is vacuous.** `np.arange(20, dtype='float16')` produces values 0..19, all with <=5 significant bits, so they encode with no `mant >>> 13` truncation, and `arrayEqual` compares stored bytes **without ever decoding**. The test passes despite all the encoder/decoder bugs above. To actually exercise the paths, use values like 65504, 0.1, inf/nan, or sub-2048 fractions, and add a decode-and-compare-to-float step.

seanmcl (commented):
Review of the fp16 support. The dtype plumbing (itemsize, Npy `f2`, byte encoders) is mostly fine, but the IEEE-754 fp16<->fp32 conversion core in `Float.lean` has several correctness bugs, and the new constructor was not fully threaded through the shared infrastructure (`isFloat`, `join`, `lossless`, `gen`, `floatVariant`). The added test is vacuous and catches none of it. Inline comments below, ranked roughly by severity.

Additional notes that didn't map cleanly to a changed line:

- **`floatVariant` (Dtype.lean:637) never produces float16.** The line-638 branch + TODO admit it should return `Dtype.float16` but returns `float32`, so `Tensor.asFloat` widens float16 tensors to float32 for all transcendental ops. Not a value-correctness bug (fp32 holds all fp16 values), but the dtype was added without finishing the one place meant to produce it.
- **Checked and NOT a bug:** `liftFloatUnop`'s `_ => throw "operation requires a float type"` looks like it rejects float16 for sin/cos/etc., but `Ufunc.asFloat` promotes float16 -> float32 first (via `floatVariant`), so the Dtype-level catch-all is never reached on the public path.
- **Minor:** the doc comment on `byteArrayToFloat16AsFloat32` (Dtype.lean:382) describes the whole arithmetic round-trip, not what the function does; `| .float16, .float16 => impossible` (line 555) could fold into the existing combined `impossible` arm; and the new code mixes Unicode `<-`-arrow style (the castOverflow / helper lines use the Unicode arrow while the surrounding file uses ASCII `<-`).

Suggested direction: rewrite the two conversion functions with explicit zero / subnormal / normal / Inf / NaN cases in both directions, round-to-nearest-even on narrowing, a corrected `0x1F` exponent test and `<<< 31` sign placement; then thread float16 through `isFloat` / `join` / `lossless` / `gen` / `floatVariant`, and replace the arange test with one that decodes non-trivial values.

TensorLib/Dtype.lean
seanmcl: I see that. It is simple and makes sense. Is it what fp16 would compute?

seanmcl:
I ran the generic `/code-review` on the PR. If something was already broken, feel free to just push back. A "TODO" comment would be great in those cases.

TensorLib/Dtype.lean
SmoothThunk: Added. I noticed it was missing while fixing order of float16 in dtype.

TensorLib/Dtype.lean
SmoothThunk: Yes. Now added details of how this gives a correct result in a comment.

TensorLib/Float.lean
SmoothThunk: Fixed.

TensorLib/Float.lean
SmoothThunk: Fixed; subnormals are encoded correctly instead of being flushed to 0.

TensorLib/Float.lean
SmoothThunk: Fixed

TensorLib/Float.lean
SmoothThunk: Implemented round-to-nearest-even in revision.

TensorLib/Float.lean
SmoothThunk: Fixed ; NaN with low payload bits won't collapse to inf

TensorLib/Float.lean
SmoothThunk: Fixed in revision

TensorLib/Npy.lean
SmoothThunk: Fixed pre-existing bug.

TensorLib/Dtype.lean
SmoothThunk: Added fp16 case so bitwise ops reject it.

TensorLib/Dtype.lean
SmoothThunk: Fixed join for float16 and int/uint.

TensorLib/Dtype.lean
SmoothThunk: Fixed; added float16 to int8 and uint8 cases.

TensorLib/Dtype.lean
SmoothThunk: Fixed; Added symmetric helpers (byteArrayTo/OfFloat16) + unwrapping versions and updated +,-,*,/, and abs. Also added roundtrip guard.

TensorLib/Dtype.lean
SmoothThunk: Fixed; float-bool calls isZero which handles -0.

TensorLib/Test.lean
SmoothThunk: Fixed; Added testFloat16EdgeCases that decodes and verifies fp16 values against expected (65504, 0.1, -0.0, inf, -inf, nan, 2048.5, 100.5). Also added PBT for roundtrip encode-decode, comm of addition, and add identity. Will include more properties based tests for other arithmetic ops soon.

TensorLib/Float.lean
```diff
@@ -96,6 +99,88 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+    else if newExp <= 0 then -- underflow for subnormal in fp16
+        -- Work with full fp32 mantissa + implicit leading 1 bit
+        -- Total shift combines fp32 to fp16 mantissa narrowing (13) + subnormal adjustment (1-newExp)
+        let totalShift := (14 - newExp).toNat
+        let fullMant := mant ||| 0x800000  -- add implicit leading 1 at bit 23
+        -- Round-to-nearest-even on discarded bits
+        let roundBit := if totalShift > 0 then (fullMant >>> (totalShift.toUInt32 - 1)) &&& 1 else 0
+        let stickyMask := if totalShift > 1 then (1 <<< (totalShift.toUInt32 - 1)) - 1 else 0
+        let stickyBits := fullMant &&& stickyMask
+        let shifted := fullMant >>> totalShift.toUInt32
```
seanmcl: **Bug (confirmed): UInt32 shift wraps modulo 32 for small fp32 values.**

In Lean 4, `UInt32.shiftRight` wraps the shift amount modulo 32: `x >>> 32 == x >>> 0 == x`. When `newExp <= -18` (fp32 biased exponent ≤ 94, i.e. |value| < 2^-33), `totalShift = (14 - newExp).toNat >= 32` and the shift wraps.

Concrete counterexample: fp32 biased exp=94, `mant=0x000001` (value ≈ 1.16e-10).
- `totalShift = 32`, `fullMant = 0x800001`
- `fullMant >>> 32` wraps to `fullMant >>> 0 = 0x800001` (should be 0)
- `rounded.toUInt16 = 0x800001 % 65536 = 1`
- Result: fp16 `0x0001` (≈ 5.96e-8) instead of ±zero

All fp32 values in this range are below the fp16 minimum subnormal (2^-24) and must encode to ±zero. The same wrap affects `roundBit` and `stickyMask` computations for `totalShift >= 33`.

Fix: cap `totalShift` at 25 (the maximum meaningful shift: 24 mantissa bits + 1 implicit bit) before any UInt32 shift, or guard `if totalShift >= 25 then sign16 else ...`.

TensorLib/Float.lean
```diff
@@ -96,6 +99,88 @@ def _root_.Float.toInt (f : Float) : Int :=
...
+      -- value = (-1)^sign x mant x 2 ^ (-24)
+      let f := Float32.ofNat mant.toNat -- mant <= 1023 fits in fp32
+      let scale := Float32.ofBits 0x33800000 -- 2 ^ (-24) in fp32
+      let result := f * scale -- shift the exponent, no precision lost
+      -- Apply sign bit
+      if sign == 1 then Float32.ofBits (result.toBits ||| 0x80000000)
+      else result
+  else if exp == 0x1F then -- exp is all 1's so this is either +-inf or NaN
+    if mant == 0 then Float32.ofBits (sign.toUInt32 <<< 31 ||| 0x7F800000) -- +- inf
+    else Float32.ofBits 0x7FC00000 -- Nan
```
seanmcl: **Behavioral divergence from NumPy: NaN decoder discards the sign bit.**

`Float32.ofBits 0x7FC00000` is returned for every fp16 NaN regardless of its sign bit. NumPy preserves the sign:
```python
np.uint16(0xFE00).view(np.float16).astype(np.float32).view(np.uint32)
# => 0xFFC00000  (sign bit 1 preserved)
```
The encoder (`Float32.toFloat16Bits`) does preserve sign on narrowing, so sign is asymmetrically handled: preserved fp32→fp16, dropped fp16→fp32.

IEEE 754 does not mandate NaN sign preservation on widening, so this is not a hard correctness error, but it is a NumPy compatibility gap. The round-trip PBT hides it via `f != f`.

Fix: `Float32.ofBits (sign.toUInt32 <<< 31 ||| 0x7FC00000)` (mirrors the Inf case on line 176).

TensorLib/Dtype.lean
```diff
@@ -866,6 +992,35 @@ example (fromDtype toDtype : Dtype) (n : Nat) :
     canCastLosslessIntRoundTrip fromDtype n toDtype
```
seanmcl: **Bug (confirmed): this PBT is now logically false for float16, but Plausible's small-Nat generator hides it.**

`lossless .float16 .float32 = true` (newly added, line 207). For `fromDtype=float16, toDtype=float32, n=3000`:
- `canCastLosslessIntRoundTrip .float16 3000 .float32` calls `float16.byteArrayOfInt 3000`
- The range check: `canCastFromInt .float16 3000` = `3000 ≤ maxSafeNatForFloat16` = `3000 ≤ 2048` = **false** → `.error`
- `canCastLosslessIntRoundTrip` returns `false`
- The property evaluates `if true then false` → counterexample

Plausible's `Arbitrary Nat` samples small values (≪ 2048 by default) so it never finds `n = 3000`. The `#guard_msgs` annotation is stale.

The semantic mismatch: `lossless` says float16→float32 loses no *value precision*, but `canCastLosslessIntRoundTrip` encodes `n` in `fromDtype` first — it conflates precision-lossless with range-within-fromDtype. Fix: scope the PBT to `n ≤ maxSafeNatForFloat16`, e.g. `example (n : Fin 2049) ...`, or document that the property tests only in-range values.

TensorLib/Dtype.lean
```diff
@@ -866,6 +992,35 @@ example (fromDtype toDtype : Dtype) (n : Nat) :
...
+---
+warning: declaration uses 'sorry'
+-/
+#guard_msgs in
+example (a : UInt16) :
+  let xa := toLEByteArray a
+  let zero := toLEByteArray (0 : UInt16)
+  let f := a.toFloat32FromFloat16
+   -- f != f is true only for NaN; check identity for everything else
+  f != f ∨ Dtype.add .float16 xa zero == .ok xa := by plausible
```
seanmcl: **Bug (confirmed): add-zero identity property fails for fp16 -0.0 (`a = 0x8000`).**

For `a = 0x8000`:
- `f = -0.0` (fp32). `-0.0 != -0.0` is `false` — the NaN escape does not fire for -0.
- `Dtype.add .float16 [0x00,0x80] [0x00,0x00]` decodes to -0.0 and +0.0, adds them. IEEE 754 round-to-nearest: **-0.0 + 0.0 = +0.0**.
- Result encodes back as `[0x00,0x00]`. But `xa = [0x00,0x80]`.
- Both disjuncts are false → genuine counterexample.

With 100 Plausible samples over 65536 UInt16 values, P(missing 0x8000 in one run) ≈ 99.85%, so the annotation usually shows "Unable to find a counter-example" but the property is wrong.

Fix options:
1. Add a second NaN-or-negative-zero escape: `(f != f || f == 0) ∨ Dtype.add .float16 xa zero == .ok xa` (since -0 and +0 are equal per IEEE, addition is identity up to sign)
2. Weaken to value equality rather than byte equality: decode both sides and compare as Float32.

TensorLib/Dtype.lean
```diff
@@ -113,6 +117,20 @@ def join (x y : Dtype) : Option Dtype :=
   let (x, y) := if x.itemsize <= y.itemsize then (x, y) else (y, x)
   if x == y then x else
   match x, y with
+  | float16, float16 => float16
```
seanmcl: **Dead arms in `join` (minor).**

Five arms here are unreachable:

1. **Line 120** `| float16, float16 => float16`: dead because `if x == y then x` at line 118 fires first for equal dtypes.
2. **Lines 129, 131, 133** `| float16, bool`, `| float16, int8`, `| float16, uint8`: dead because after the size-normalization swap (`if x.itemsize <= y.itemsize`), float16 (size 2) is always sorted to the *y* side when paired with size-1 types. The correct results already fall through to `| bool, _ => y`, `| int8, float16 => float16`, and `| uint8, float16 => float16`.

No wrong output is produced, but the dead arms are misleading. Consider removing them.

TensorLib/Dtype.lean
```diff
@@ -328,6 +357,12 @@ def byteArrayOfFloat32 (dtype : Dtype) (f : Float32) : Err ByteArray := match dt
 
 private def byteArrayOfFloat32! (dtype : Dtype) (f : Float32) : ByteArray := get! $ byteArrayOfFloat32 dtype f
 
+def byteArrayOfFloat16 (dtype : Dtype) (f : Float32) : Err ByteArray :=
+  if dtype == .float16 then .ok (toLEByteArray f.toFloat16Bits)
+  else .error "Illegal type conversion"
+
+private def byteArrayOfFloat16! (dtype : Dtype) (f : Float32) : ByteArray := get! $ byteArrayOfFloat16 dtype f
```
seanmcl: **`byteArrayOfFloat16!` is unused dead code.** No call sites exist anywhere in the repo — grep confirms zero hits. The public `byteArrayOfFloat16` is called directly. The `!`-suffix variant is compiled but unreachable; consider removing it.

TensorLib/Dtype.lean
```diff
@@ -362,6 +397,36 @@ private def byteArrayToFloat32RoundTrip (dtype : Dtype) (f : Float32) : Bool :=
 #guard float32.byteArrayToFloat32RoundTrip (Float32.sqrt 2)
 #guard float32.byteArrayToFloat32RoundTrip 0
 
+-- Read 2 bytes as fp 16, upcast to fp32 for computations
+def byteArrayToFloat16 (dtype : Dtype)(arr : ByteArray) : Err Float32 := match dtype with
+  | .float16 => arr.toUInt16LE.map UInt16.toFloat32FromFloat16
+  | _ => .error "Illegal type conversion"
+
+def byteArrayToFloat16! (dtype : Dtype) (arr : ByteArray) : Float32 := get! $ byteArrayToFloat16 dtype arr
```
seanmcl: **`byteArrayToFloat16!` is unused dead code.** Same situation — no call sites. Also, `castOverflow` and `isZero` still call the older `byteArrayToFloat16AsFloat32` helper rather than the new dtype-validated `byteArrayToFloat16`. If `byteArrayToFloat16AsFloat32` is intended to be retired, those two call sites need updating; if it stays, the `!` variant is unnecessary noise.

seanmcl (commented):
Review of the revision. The conversion rewrites fix the headline bugs from the first round (correct `0x1F` Inf/NaN test, `<<< 31` sign placement, subnormal decode, NaN payload preservation, `isFloat`/`gen`/`floatVariant`/`lossless`/`join` threading). Three real correctness issues remain, plus a few minor items.

**Summary of remaining issues:**
1. **UInt32 shift wraparound** in the subnormal fp32→fp16 encoder (confirmed concrete counterexample)
2. **Two property-based tests are wrong** — the lossless round-trip PBT is logically falsified by float16, and the add-zero identity PBT fails for fp16 -0.0
3. **NaN decoder drops sign bit** (behavioral divergence from NumPy)
4. Several dead arms in `join` and an unused `!`-variant helper

TensorLib/Float.lean
SmoothThunk: Fixed; added guard as mentioned for totalShift >= 25 to return +-0. Added test cases for 1.16e-10, 1e-20, -1e-20, and 1e-40 to verify.

TensorLib/Float.lean
SmoothThunk: Fixed so NaN decoder preserves sign bit

TensorLib/Dtype.lean
SmoothThunk: Fixed; added guard to skip values outside fromDtypes representable range so the PBT only tests the roundtrip for values that are in range for given Dtype.

TensorLib/Dtype.lean
SmoothThunk: Fixed PBT using (1) if the value is +-0, skip the byte equality check.

TensorLib/Dtype.lean
SmoothThunk: Replaced with fp16, _ => fp16.

TensorLib/Dtype.lean
SmoothThunk: Removed

TensorLib/Dtype.lean
SmoothThunk: Fixed; removed byteArrayToFloat16! and byteArrayToFloat16AsFloat32. castOverflow and isZero now call byteArrayToFloat16.

seanmcl (commented):
Review of revision 3. The fixes from the prior round all landed correctly:
- UInt32 shift-wraparound guarded at `totalShift >= 25` ✓
- NaN decoder preserves sign bit ✓  
- Add-zero PBT extended with `-0` escape (`f == 0`) ✓
- Lossless PBT range-guarded with `maxSafeNat` ✓
- Dead `join` arms (float16/float16, float16/bool, etc.) removed ✓
- `byteArrayToFloat16AsFloat32`, `byteArrayToFloat16!`, `byteArrayOfFloat16!` all removed ✓
- Tests extended with below-subnormal values ✓

Two remaining issues:

**1. (Bug) Self-cast PBT still unguarded** — the pre-existing property `example (dtype : Dtype) (n : Nat) : canCastLosslessIntRoundTrip dtype n dtype` (Dtype.lean line 965, inside the `Test` section, not in this diff) is now falsified by float16 being added to `gen`. For `dtype = float16, n = 2049`: `byteArrayOfInt .float16 2049` fails the range check (`2049 > maxSafeNatForFloat16 = 2048`), so `canCastLosslessIntRoundTrip` returns `false` — a genuine counterexample. The lossless PBT at line 974 got the `maxSafeNat` guard; this one didn't. Plausible's Nat generator uses small sizes by default and won't find 2049, but the property is logically wrong. Fix: add the same guard — `if n > dtype.maxSafeNat.getD n then true else canCastLosslessIntRoundTrip dtype n dtype`.

**2. (Minor) Dead catch-all arm in `join`** — see inline comment.

The implementation itself looks correct. The `totalShift >= 25` guard is right (verified: the rounding at totalShift=24 produces correct IEEE RNE behavior), and the NaN sign fix is consistent with NumPy.

TensorLib/Dtype.lean
```diff
@@ -113,7 +117,16 @@ def join (x y : Dtype) : Option Dtype :=
...
+  | float16, float32 => float32
+  | float16, float64 => float64
+  | float16, int16 => float32
+  | float16, uint16 => float32
+  | float16, int32 => float64
+  | float16, uint32 => float64
+  | float16, int64 => float64
+  | float16, uint64 => float64
   | float32, float64 => float64
+  | float16, _ => float16 -- for all other cases that promote to fp16
```
seanmcl: **Minor: this catch-all is unreachable dead code.**

After the normalization `let (x, y) := if x.itemsize <= y.itemsize then ...`, `x = float16` only when `y.itemsize >= 2`. All such y values are explicitly listed in the eight arms above: float32, float64, int16, uint16, int32, uint32, int64, uint64. The case `y = float16` is caught by `if x == y then x` before the match. There is no dtype with itemsize ≥ 2 left for this arm to match.

The comment `-- for all other cases that promote to fp16` is misleading — there are no such cases with the current set of dtypes. If a future dtype with itemsize 2 (e.g. bfloat16) is added, this arm would silently return `float16` rather than requiring an explicit decision. Consider either removing it or replacing it with an explicit `impossible` to document the intent.

TensorLib/Dtype.lean
```diff
@@ -862,10 +974,42 @@ warning: declaration uses 'sorry'
 #guard_msgs in
 -- Lossless translations should be OK
 example (fromDtype toDtype : Dtype) (n : Nat) :
+-- only try values within fromDtye's representable range
+-- fp16 max n = 2^11, fp32 max n = 2^24, fp64 = 2^53, uint8 = 255, int8 = 127, etc
   if lossless fromDtype toDtype then
-    canCastLosslessIntRoundTrip fromDtype n toDtype
+    if n > fromDtype.maxSafeNat.getD n then true else canCastLosslessIntRoundTrip fromDtype n toDtype
```
seanmcl, line 985: **Bug (confirmed): the self-cast PBT at line 965 (`canCastLosslessIntRoundTrip dtype n dtype`) is now falsified by float16 in `gen`, but Plausible's Nat generator masks it.**

This range guard correctly fixes the *lossless* PBT. However the earlier property (a few lines up):
```lean
-- One dtype should always go back and forth
example (dtype : Dtype) (n : Nat) :
  canCastLosslessIntRoundTrip dtype n dtype := by plausible
```
has no such guard. With float16 now in `gen`, sampling `dtype = float16, n = 2049`: `byteArrayOfInt .float16 2049` checks `canCastFromInt float16 2049` = `2049 ≤ 2048` = **false** → `.error` → `canCastLosslessIntRoundTrip` returns `false`. The property fires `false`. Plausible generates small Nat values (≪ 2048) so it never finds `n = 2049`, but the property is logically wrong.

The same failure exists for int8 (n > 127), uint8 (n > 255), int16/uint16 (n > 32767) — these were already broken before this PR, but float16 adds a smaller bound (2048) that's easier to hit.

Fix: apply the same guard: `if n > dtype.maxSafeNat.getD n then true else canCastLosslessIntRoundTrip dtype n dtype`.

seanmcl (approved):

seanmcl:
Would good to tal at the last couple comments, but looks solid now.

TensorLib/Dtype.lean
SmoothThunk: Replaced | fp16, _ => impossible for now. I'm unable to remove the line of code since the exhaustiveness checker gives me missing case errors.

TensorLib/Dtype.lean
SmoothThunk: Fixed; added a range guard that skips values outside given dtypes range.

TensorLib/Dtype.lean
SmoothThunk: Updated fix uses join recursively which avoids us having to write unreachable arms inside match.

PR #95, opened 2026-07-15
Add float8_e4m3fn dtype support

SmoothThunk:
Fp8_e4m3 = 1 sign, 4 exp, 3 mantissa (bias = 7, max + +-448, no inf, NaN only)
1. Encoder uses round to nearest even, overflow / inf -> NaN as per ml_dtypes
2. Decoder handles 0, subnormals, normals, NaN
3. Pattern matches have been updated with cases for this fp8 type
4. Npy parsing includes <V1 descriptor
5. Includes testing and PBT for round trip property and add commutativity.

### DIFF
--- a/TensorLib/Dtype.lean
+++ b/TensorLib/Dtype.lean
@@ -146,6 +146,8 @@ def join (x y : Dtype) : Option Dtype :=
       | .float8_e4m3, .float32 => float32
       | .float8_e4m3, .float64 => float64
       | .float8_e4m3, .bool
+      -- numpy promotes int8/uint8 + fp8_e4m3 => fp8_e4m3 (lossy for values > 16, e.g. 100 -> 96)
+      -- matches ml_dtypes behavior; lossless correctly returns false for int8 to fp8_e4m3
       | .float8_e4m3, .int8
       | .float8_e4m3, .uint8 => float8_e4m3
       | .float8_e4m3, _ => none
@@ -634,6 +636,8 @@ def abs (dtype : Dtype) (x : ByteArray) : Err ByteArray := do
   | .uint8 | .uint16 | .uint32 | .uint64 => return x
   | .int8 | .int16| .int32 | .int64 => return dtype.byteArrayOfIntOverflow x.toInt.natAbs
   | .float8_e4m3 => do
+    -- prevent malformed / empty input since abs has no size check
+    if x.size != 1 then .error "abs : byte size mismatch" else
     let x := x.data[0]!.toFloat32FromFloat8E4M3
     return ByteArray.mk #[x.abs.toFloat8E4M3Bits]
   | .float16
@@ -647,6 +651,10 @@ def abs (dtype : Dtype) (x : ByteArray) : Err ByteArray := do
     let x <- dtype.byteArrayToFloat64 x
     dtype.byteArrayOfFloat64 x.abs
   | .bool => return x
+
+-- abs rejects wrong-sized input for fp8_e4m3
+#guard (Dtype.abs .float8_e4m3 (ByteArray.mk #[])).isOk == false -- 0 bytes (should be atleast 1 byte for fp8_e4m3)
+#guard (Dtype.abs .float8_e4m3 (ByteArray.mk #[1, 2])).isOk == false -- 2 bytes (too many for a 1 byte type)
 
 def abs! (dtype : Dtype) (x : ByteArray) : ByteArray := get! $ abs dtype x
 
@@ -663,6 +671,7 @@ def isZero (dtype : Dtype) (x : ByteArray) : Err Bool := match dtype with
 | uint64 => return x.data.all fun v => v == 0
 -- We need to worry about -0, which is not all 0s in the bit pattern.
 | float8_e4m3 => do
+  if x.size != 1 then .error "isZero: byte size mismatch" else
   let f := x.data[0]!.toFloat32FromFloat8E4M3
   return f == 0
 | float16
@@ -675,6 +684,10 @@ def isZero (dtype : Dtype) (x : ByteArray) : Err Bool := match dtype with
 | float64 => do
   let f <- Float.ofLEByteArray x
   return f == 0
+
+-- isZero rejects wrong-sized input for float8_e4m3
+#guard (Dtype.isZero .float8_e4m3 (ByteArray.mk #[])).isOk == false
+#guard (Dtype.isZero .float8_e4m3 (ByteArray.mk #[1, 2])).isOk == false
 
 def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err ByteArray :=
   if fromDtype == toDtype then return data else
@@ -741,22 +754,27 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err 
       encodeFloat16OrBFloat16 toDtype f
     -- float8_e4m3 to unsigned integers
     | .float8_e4m3, .uint8 | .float8_e4m3, .uint16 | .float8_e4m3, .uint32 | .float8_e4m3, .uint64 => do
+      if data.size != 1 then .error "castOverflow: byte size mismatch" else
       let f := data.data[0]!.toFloat32FromFloat8E4M3
       return toDtype.byteArrayOfNatOverflow f.toNat
     -- float8_e4m3 to signed integers
     | .float8_e4m3, .int8 | .float8_e4m3, .int16 | .float8_e4m3, .int32 | .float8_e4m3, .int64 => do
+      if data.size != 1 then .error "castOverflow: byte size mismatch" else
       let f := data.data[0]!.toFloat32FromFloat8E4M3
       return toDtype.byteArrayOfIntOverflow f.toInt
     -- float8_e4m3 to float32
     | .float8_e4m3, .float32 => do
+      if data.size != 1 then .error "castOverflow: byte size mismatch" else
       let f := data.data[0]!.toFloat32FromFloat8E4M3
       return toLEByteArray f
     -- float8_e4m3 to float64
     | .float8_e4m3, .float64 => do
+      if data.size != 1 then .error "castOverflow: byte size mismatch" else
       let f := data.data[0]!.toFloat32FromFloat8E4M3
       return toLEByteArray f.toFloat
     -- float8_e4m3 to fp16/bf16
     | .float8_e4m3, .float16 | .float8_e4m3, .bfloat16 => do
+      if data.size != 1 then .error "castOverflow: byte size mismatch" else
       let f := data.data[0]!.toFloat32FromFloat8E4M3
       encodeFloat16OrBFloat16 toDtype f
     -- float32 -> float8_e4m3
--- a/TensorLib/Dtype.lean
+++ b/TensorLib/Dtype.lean
@@ -333,6 +333,16 @@ private def canCastFromInt (dtype : Dtype) (n : Int) : Bool :=
 
 def sizedStrides (dtype : Dtype) (s : Shape) : Strides := List.map (fun x => x * dtype.itemsize) s.unitStrides
 
+-- Decode 1-byte fp8_e4m3 to fp32
+-- Prevent having to repeat size check at call sites
+def decodeFloat8E4M3 (arr : ByteArray) : Err Float32 :=
+  if arr.size != 1 then .error "decoder: expected 1 byte for float8_e4m3"
+  else .ok (arr.data[0]!.toFloat32FromFloat8E4M3)
+
+-- Encode Float32 to 1-byte fp8_e4m3
+def encodeFloat8E4M3 (f : Float32) : ByteArray :=
+  ByteArray.mk #[f.toFloat8E4M3Bits]
+
 def byteArrayOfNatOverflow (dtype : Dtype) (n : Nat) : ByteArray := match dtype with
 | .bool => toLEByteArray (if n == 0 then 0 else 1).toUInt8
 | .uint8 => toLEByteArray n.toUInt8
@@ -343,7 +353,7 @@ def byteArrayOfNatOverflow (dtype : Dtype) (n : Nat) : ByteArray := match dtype 
 | .int32 => toLEByteArray n.toInt32
 | .uint64 => toLEByteArray n.toUInt64
 | .int64 => toLEByteArray n.toInt64
-| .float8_e4m3 => toLEByteArray n.toFloat32.toFloat8E4M3Bits
+| .float8_e4m3 => encodeFloat8E4M3 n.toFloat32
 | .float16 => toLEByteArray n.toFloat32.toFloat16Bits
 | .bfloat16 => toLEByteArray n.toFloat32.toBFloat16Bits
 | .float32 => toLEByteArray n.toFloat32
@@ -373,7 +383,7 @@ private def byteArrayOfIntOverflow (dtype : Dtype) (n : Int) : ByteArray := matc
 | .uint16 | .int16 => toLEByteArray n.toInt16
 | .uint32 | .int32 => toLEByteArray n.toInt32
 | .uint64 | .int64 => toLEByteArray n.toInt64
-| .float8_e4m3 => toLEByteArray n.toFloat32.toFloat8E4M3Bits
+| .float8_e4m3 => encodeFloat8E4M3 n.toFloat32
 | .float16 => toLEByteArray n.toFloat32.toFloat16Bits
 | .bfloat16 => toLEByteArray n.toFloat32.toBFloat16Bits
 | .float32 => toLEByteArray n.toFloat32
@@ -480,7 +490,6 @@ private def byteArrayToFloat16RoundTrip (dtype : Dtype) (f : Float32) : Bool :=
 #guard float16.byteArrayToFloat16RoundTrip 42
 #guard float16.byteArrayToFloat16RoundTrip (-0)
 
-
 -- Decode bf16 bytes to Float32. bf16 shares fp32's exponent, so just pad mantissa with 0's.
 def byteArrayToBFloat16 (dtype : Dtype) (arr : ByteArray) : Err Float32 := match dtype with
   | .bfloat16 => arr.toUInt16LE.map UInt16.toFloat32FromBFloat16
@@ -522,9 +531,9 @@ def add (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
   | .int8 | .int16| .int32 | .int64 => do
     dtype.byteArrayOfInt (x.toInt + y.toInt)
   | .float8_e4m3 => do
-    let x := x.data[0]!.toFloat32FromFloat8E4M3
-    let y := y.data[0]!.toFloat32FromFloat8E4M3
-    return ByteArray.mk #[(x + y).toFloat8E4M3Bits]
+    let x <- decodeFloat8E4M3 x
+    let y <- decodeFloat8E4M3 y
+    return encodeFloat8E4M3 (x + y)
   | .float16
   | .bfloat16 => do
     let x <- dtype.decodeFloat16OrBFloat16 x
@@ -549,9 +558,9 @@ def sub (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
   | .int8 | .int16| .int32 | .int64 => do
     return dtype.byteArrayOfIntOverflow (x.toInt - y.toInt)
   | .float8_e4m3 => do
-    let x := x.data[0]!.toFloat32FromFloat8E4M3
-    let y := y.data[0]!.toFloat32FromFloat8E4M3
-    return ByteArray.mk #[(x - y).toFloat8E4M3Bits]
+    let x <- decodeFloat8E4M3 x
+    let y <- decodeFloat8E4M3 y
+    return encodeFloat8E4M3 (x - y)
   | .float16
   | .bfloat16 => do
     let x <- dtype.decodeFloat16OrBFloat16 x
@@ -577,9 +586,9 @@ def mul (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
   | .int8 | .int16| .int32 | .int64 => do
     return dtype.byteArrayOfIntOverflow (x.toInt * y.toInt)
   | .float8_e4m3 => do
-    let x := x.data[0]!.toFloat32FromFloat8E4M3
-    let y := y.data[0]!.toFloat32FromFloat8E4M3
-    return ByteArray.mk #[(x * y).toFloat8E4M3Bits]
+    let x <- decodeFloat8E4M3 x
+    let y <- decodeFloat8E4M3 y
+    return encodeFloat8E4M3 (x * y)
   | .float16
   | .bfloat16 => do
     let x <- dtype.decodeFloat16OrBFloat16 x
@@ -605,9 +614,9 @@ def div (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
   | .int8 | .int16| .int32 | .int64 => do
     return dtype.byteArrayOfIntOverflow (x.toInt / y.toInt)
   | .float8_e4m3 => do
-    let x := x.data[0]!.toFloat32FromFloat8E4M3
-    let y := y.data[0]!.toFloat32FromFloat8E4M3
-    return ByteArray.mk #[(x / y).toFloat8E4M3Bits]
+    let x <- decodeFloat8E4M3 x
+    let y <- decodeFloat8E4M3 y
+    return encodeFloat8E4M3 (x / y)
   | .float16
   | .bfloat16 => do
     let x <- dtype.decodeFloat16OrBFloat16 x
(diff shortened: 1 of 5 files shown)


SmoothThunk:
Code review findings on this PR:

### 1. `join` commutativity regression — `TensorLib/Dtype.lean:148`

**Summary:** `join` is not commutative between `float8_e4m3` and `bool` / `int8` / `uint8`. All four dtypes have `itemsize == 1`, so the swap guard `x.itemsize > y.itemsize` never triggers, and the `float8_e4m3`-first arm returns `none` while the other-first arms return `float8_e4m3`.

**Failure scenario:**
- `Dtype.join .uint8 .float8_e4m3` matches `.uint8, _ => y` → `some .float8_e4m3`
- `Dtype.join .float8_e4m3 .uint8` matches `.float8_e4m3, _ => none`

Same asymmetry for `.bool` (`.bool, _ => y` vs `.float8_e4m3, _ => none`) and `.int8` (`.int8, _ => y` vs `.float8_e4m3, _ => none`). This is the same class of bug that was fixed for `int64`/`uint64` in commit 3489aee. Any downstream op that relies on `join` for type promotion now depends on argument order.

### 2. `isFloat` includes `float8_e4m3` but `liftFloatUnop` does not — `TensorLib/Dtype.lean:110`

**Summary:** `isFloat` now returns `true` for `float8_e4m3`, but `liftFloatUnop` (which powers `arctan` / `cos` / `exp` / `log` / `sin` / `sqrt` / `tan` / `tanh`) only handles `float16` / `bfloat16` / `float32` / `float64` and throws `"operation requires a float type"` on `float8_e4m3`.

**Failure scenario:** A caller that gates on `dtype.isFloat` before dispatching to `Dtype.exp` / `Dtype.sin` / etc. — a natural pattern given the name — will call the op on a `float8_e4m3` ByteArray and get the runtime error `"operation requires a float type"`. This also breaks the invariant that `isFloat` predicts which dtypes support the transcendental ops (which was the stated reason `bfloat16` was added there originally, per the comment above `isFloat`).

### 3. Dead branch in encoder — `TensorLib/Float.lean:322`

**Summary:** The `else if e4m3Exp + 1 > 15 then (15, (7 : UInt32))` branch is unreachable inside the surrounding `realExp >= -6 && realExp <= 7` arm.

**Failure scenario:** With `realExp ∈ [-6, 7]`, `e4m3Exp = realExp + 7 ∈ [1, 14]`, so `e4m3Exp + 1 ∈ [2, 15]` and the `> 15` case never fires. The branch (plus its `-- Would overflow — but this shouldn't happen in normal range` comment) can be removed; the block simplifies to `(e4m3Exp + 1, (0 : UInt32))`, with the subsequent `finalExp == 15 && finalMant == 7` NaN check handling the remaining edge case.

SmoothThunk:
## Fixes:
* Added explicit | .float8_e4m3, .bool => float8_e4m3, | .float8_e4m3, .int8 => float8_e4m3, | .float8_e4m3, .uint8 => float8_e4m3 before the catch-all. Both directions now correctly return float8_e4m3. The catch all is still needed for int16, uint16, int32, etc where it can't be promoted.
* Added float8_e4m3 case to liftFloatUnop.
* Removed the dead branch since e4m3Exp is in [1, 14] in this path, e4m3Exp + 1 can never exceed 15.

SmoothThunk:
## Self-review at max effort — float8_e4m3 branch

Six findings, most-severe first:

### 1. `Dtype.abs` fp8 branch panics on wrong-sized input (`TensorLib/Dtype.lean:637`)
Uses `x.data[0]!` with no size guard. Every other float branch in `abs` returns `.error`, so fp8 is uniquely crash-prone. `abs` is one of the few ops without an upfront `itemsize != x.size` check, so a malformed/empty ByteArray reaches this branch and panics.

### 2. `Dtype.isZero` fp8 branch same panic (`TensorLib/Dtype.lean:666`)
`x.data[0]!` with no guard. The `.float16`/`.bfloat16`/`.float32`/`.float64` branches all return `Err`. Reachable via `castOverflow`'s float→bool path (`castOverflow` calls `fromDtype.isZero data` for any float source), so a malformed fp8 buffer cast to bool crashes.

### 3. `join .float8_e4m3 .int8/.uint8 → .float8_e4m3` is lossy (`TensorLib/Dtype.lean:149`)
fp8_e4m3 has only 4 significant bits (implicit 1 + 3 mantissa), so `int8` `100` rounds to `96` in fp8. The file's own comment says this models NumPy's binary-op promotion, which promotes small-int + small-float to at least float16/float32 (e.g. `int8 + float16 → float32`). Any downstream consumer relying on the joined dtype to represent the integer side losslessly gets silent precision loss.

### 4. Encoder overflow boundary is non-monotonic (`TensorLib/Float.lean:286`)
The `realExp > 7` branch handles `realExp == 8`. Take fp32 `464.0` (halfway between 448 and unrepresentable 480): `truncated = 14`, `roundBit = 1`, `stickyBits = 0`, `truncated & 1 = 0` → no round-up → `rounded = 14` → `mantBits = 6` → encoded as `126` = 448.0. But fp32 `465.0` has `stickyBits ≠ 0` → `rounded = 15` → `mantBits = 7` → NaN. So 464 → 448 but 465 → NaN: values 1 fp32-ULP apart flip from finite-max to NaN, and finite 464 (larger than the max representable 448) silently rounds down instead of saturating to NaN. `ml_dtypes` e4m3fn treats every value above 448 (absolute) as overflow → NaN.

### 5. `castOverflow` fp8-source branches panic on short input (`TensorLib/Dtype.lean:743–761`)
Five branches use `data.data[0]!` with no size guard. The analogous fp16/bf16 branches use `decodeFloat16OrBFloat16` (returns `.error` on size mismatch); the fp32/fp64 branches use `Float32.ofLEByteArray`/`Float.ofLEByteArray` (also `.error`). So a caller feeding an untrusted/misparsed byte slice into a cast gets a crash only when the source dtype is fp8.

### 6. Encoder maps negative NaN to `0x7F`, not `0xFF` (`TensorLib/Float.lean:262`)
Lean's `Float32.toBits` normalizes NaN sign, so all NaN inputs encode as `+NaN` (0x7F). This diverges from `ml_dtypes`, which preserves NaN sign. The `toFloat16Bits`/`toBFloat16Bits` docstrings explicitly document this divergence (lines 107, 163–164); the new fp8 encoder's docstring instead writes "0x7F / 0xFF" as if sign were preserved. Either document the divergence or fix the sign propagation for NaN inputs.

### Suggested minimum fixes
- Add `if x.size != dtype.itemsize` guards at the top of `abs`, `isZero`, and each fp8-source branch of `castOverflow` — or route fp8 through a safe decoder that returns `Err` (mirroring `decodeFloat16OrBFloat16`).
- Reconsider `join` for fp8 × int8/uint8/bool: NumPy promotes to a type large enough for both operands.
- Decide the saturation rule at 464: either saturate everything above 448 to NaN (match ml_dtypes) or saturate to 448 — but not the current asymmetric behavior.
- Update the fp8 encoder docstring to note that negative NaN inputs encode as `+NaN` due to Lean's NaN normalization, matching the fp16/bf16 comments.

SmoothThunk:
## Fixes:

* Added size guard if x.size != 1 before accessing x.data[0]! so it returns .error instead of panicking on bad input. Added #guard tests for empty and oversized ByteArrays.
* Added size guard to isZero for fp8_e4m3, same pattern as abs. Added #guard tests for empty and oversized input.
* Unchanged since join in npy returns fp8_e4m3 so our join matches that. The lossless function separately returns false for int8, fp8_e4m3 which is correct.
* In ml_dtypes, values 449–464 become 448 (bits=126), and values >= 465 become NaN. The encoder matches this behavior exactly. Added test cases for both.
* Added size guard to fp8, _ branches in castOverflow so they error on bad input.
* Updated the comment to document NaN sign divergence. Sign is preserved for +-inf but not for NaN due to Lean's normalization.

SmoothThunk:
## Code-review findings

**TensorLib/Dtype.lean:141** — the `.float16, .float8_e4m3 => none` match arm in `join` is unreachable dead code.

Because `join` swaps arguments when `x.itemsize > y.itemsize` (and `float16` is 2 bytes vs. `float8_e4m3` at 1 byte), any `join .float16 .float8_e4m3` call recurses as `join .float8_e4m3 .float16`, which is handled by the `.float8_e4m3, _ => none` catch-all at line 153. The line can be removed, leaving just `| .float16, .bfloat16 => none`.

SmoothThunk:
Removed dead arm in join.

TensorLib/Float.lean
```diff
@@ -220,6 +222,134 @@ def _root_.UInt16.toFloat32FromFloat16 (bits : UInt16) : Float32 :=
 def _root_.UInt16.toFloat32FromBFloat16 (bits : UInt16) : Float32 :=
   Float32.ofBits (bits.toUInt32 <<< 16)
 
+-- Convert float8_e4m3fn bits (stored as UInt8) to Float32.
+-- E4M3FN format: 1 sign bit + 4 exponent bits + 3 mantissa bits, bias = 7
+-- Special values: No infinity. Exponent all 1s (0xF) with any mantissa = NaN.
```
seanmcl: This comment is inaccurate for e4m3**fn** and contradicts the code below it. Only `exp==0xF && mant==0x7` is NaN (line 249); `exp==0xF` with mant `0..6` are **normal** values 256, 288, …, 448 (byte `0x7E` = 448, exercised by the `126 -> 0x43E00000` guard). Line 228 already says the right thing ("only two NaN encodings since mant must be 0x7"), so line 227 contradicts it. The risk: someone trusting "any mantissa = NaN" could simplify the guard to `exp == 0xF` and silently turn ±256..±448 (14 encodings) into NaN. Suggest: "Exponent all 1s (0xF) with mantissa 0x7 = NaN; mantissa 0..6 are normal (up to 448)."

TensorLib/Float.lean
```diff
@@ -220,6 +222,134 @@ def _root_.UInt16.toFloat32FromFloat16 (bits : UInt16) : Float32 :=
...
+    else
+      -- Subnormal: no implicit leading 1
+      -- value = (-1)^sign × mant × 2^(1 - 7 - 3) = mant × 2^(-9)
+      let f := Float32.ofNat mant.toNat
+      let scale := Float32.ofBits 0x3B000000  -- 2^(-9) in fp32
+      let result := f * scale
+      if sign == 1 then Float32.ofBits (result.toBits ||| 0x80000000) else result
+  else if exp == 0xF && mant == 0x7 then
+    -- Exponent all 1s: NaN (e4m3fn has NO infinity, only NaN)
+    -- We map all NaN bit patterns to fp32 quiet NaN, preserving sign
```
seanmcl: Minor: this says NaN sign is "preserving sign", but the encode function's comment (line ~263) correctly states the opposite — Lean's `Float32` doesn't round-trip a NaN sign bit, so byte `0xFF` (-NaN) doesn't come back out as a negative NaN, diverging from ml_dtypes. No test checks NaN sign (they use `f != f`), so this is harmless, but the two comments should agree — recommend matching the honest wording used at line 263.

TensorLib/Float.lean
```diff
@@ -220,6 +222,134 @@ def _root_.UInt16.toFloat32FromFloat16 (bits : UInt16) : Float32 :=
...
+      let truncated := mant >>> 20
+      let roundBit := (mant >>> 19) &&& 1
+      let stickyBits := mant &&& 0x7FFFF
+      let rounded := if roundBit == 1 && (stickyBits != 0 || truncated &&& 1 == 1)
+        then truncated + 1 else truncated
+      -- If rounding overflows mantissa (0b1000), bump exponent
+      -- e4m3Exp is [1, 14] here so e4m3Exp + 1 is always in [2, 15] and never overflows.
+      let (finalExp, finalMant) := if rounded > 0x7 then (e4m3Exp + 1, (0 : UInt32)) else (e4m3Exp, rounded)
+      -- Check if we accidentally hit NaN pattern (exp=15, mant=7)
+      if finalExp == 15 && finalMant == 7 then
```
seanmcl: Nit (dead code): in this branch `realExp ∈ [-6,7]` ⇒ `e4m3Exp ∈ [1,14]`, and `finalExp` only reaches 15 via the mantissa carry (`rounded > 0x7`), which forces `finalMant = 0`. So `finalExp == 15 && finalMant == 7` can never both hold — I instrumented 5M random inputs and it fires 0 times. It's harmless but can be dropped (the reachable overflow→NaN is handled by the `realExp > 7` branch). Not blocking.

seanmcl (commented):
Thanks for this — the core fp8 conversion is really solid. I verified the encoder/decoder against `ml_dtypes.float8_e4m3fn` and found **0 mismatches** across all 256 decode bytes, ~2M random fp32 encodes, all 17.6k rounding-tie midpoints, and all 65,536 operand pairs for add/sub/mul/div. Every `#guard`/test constant checks out, the `join` promotion table matches `np.result_type`, and `castOverflow` is exhaustive. Nice work on the round-to-nearest-even handling.

The issues I found are all at the integration edges, not the bit math. Inline comments below; one that can't be inlined (the file isn't in this diff):

**🔴 `Tensor.lean:602` — `toFloat32Tree`/`toFloat64Tree` don't handle fp8.** These dispatch fp16/bf16 explicitly (lines 600-601) but route everything else through `| _ => Float32.ofLEByteArray b` / `Float.ofLEByteArray b`, which require exactly 4/8 bytes. On a 1-byte fp8 element the `Err` variant returns `"Expected size 4 byte array"` for every element, and the `!` variant panics. fp8 tensors can't be converted/formatted to float trees at all — a functional gap vs fp16/bf16. Worth adding a `.float8_e4m3` arm (this mirrors the same class of miss as adding an enum case without widening a `_` dispatch).

TensorLib/Npy.lean
```diff
@@ -110,6 +110,7 @@ def dtypeNameToNpyString (t : TensorLib.Dtype) : String := match t with
 | .uint16 => "u2"
 | .uint32 => "u4"
 | .uint64 => "u8"
+| .float8_e4m3 => "V1"
```
seanmcl, line 113: Heads-up (not necessarily actionable here): `<V1` is not unique to e4m3fn — `ml_dtypes` serializes `float8_e4m3fnuz`, `float8_e5m2fnuz`, `float8_e3m4`, `float8_e4m3b11fnuz`, and `float8_e8m0fnu` all with descr `<V1` too (only `float8_e5m2` uses `<f1`). So a valid `.npy` of one of those types would be silently read as e4m3fn with wrong values. This is the same aliasing tradeoff already accepted for bf16's `<V2`, but fp8 has many more colliders — might be worth a comment noting the assumption, like the bf16 branch has.

TensorLib/Npy.lean
```diff
@@ -124,6 +125,7 @@ def fromNpyString (s : String) : Err Dtype :=
     -- We only recognize "<V2" as bf16 to avoid collision with "|V2" (actual void data).
     -- Only littleEndian V2 is bf16. Tensor.toNpy always writes LE so this is safe.
     let name <- if nameStr == "V2" && order == .littleEndian then .ok .bfloat16
+      else if nameStr == "V1" && order == .littleEndian then .ok .float8_e4m3
```
seanmcl, line 130: The `<V2`→bf16 branch has explicit round-trip/collision guards (`Npy.lean:427-428`: `<V2` → bf16, `|V2` ≠ bf16). The new `<V1`→fp8 branch has no equivalent test. Since correctness here rests entirely on the `&& order == .littleEndian` discriminator (to avoid genuine `|V1` numpy void), a regression dropping that guard wouldn't be caught. Suggest adding the parallel guards:
```lean
#guard Npy.Dtype.fromNpyString "<V1" == .ok { name := .float8_e4m3, order := .littleEndian }
#guard Npy.Dtype.fromNpyString "|V1" != .ok { name := .float8_e4m3, order := .littleEndian }
```

TensorLib/Dtype.lean
```diff
@@ -500,6 +521,10 @@ def add (dtype : Dtype) (x y : ByteArray) : Err ByteArray :=
     return dtype.byteArrayOfNatOverflow (x.toNat + y.toNat)
   | .int8 | .int16| .int32 | .int64 => do
     dtype.byteArrayOfInt (x.toInt + y.toInt)
+  | .float8_e4m3 => do
```
seanmcl, line 533: Reuse nit: the fp8 arms in `add`/`sub`/`mul`/`div`/`abs`/`isZero`/`liftFloatUnop`/`castOverflow` open-code `x.data[0]!.toFloat32FromFloat8E4M3` (decode) and `ByteArray.mk #[….toFloat8E4M3Bits]` (encode) at ~25 sites, whereas fp16/bf16 funnel through `decodeFloat16OrBFloat16`/`encodeFloat16OrBFloat16` and collapse to one shared arm. A `decodeFloat8E4M3 : ByteArray → Err Float32` / `encodeFloat8E4M3` pair would remove the duplication and also make the size check uniform (right now `abs`/`isZero`/`castOverflow` guard `size != 1` inline, while add/sub/mul/div rely on the function-head `itemsize` guard — safe today, but the invariant is enforced three different ways). Not blocking.

TensorLib/Dtype.lean
```diff
@@ -697,7 +751,44 @@ def castOverflow (fromDtype : Dtype) (data : ByteArray) (toDtype : Dtype) : Err
...
+      if data.size != 1 then .error "castOverflow: byte size mismatch" else
+      let f := data.data[0]!.toFloat32FromFloat8E4M3
+      encodeFloat16OrBFloat16 toDtype f
+    -- float32 -> float8_e4m3
+    | .float32, .float8_e4m3 => do
+      let f <- Float32.ofLEByteArray data
+      return ByteArray.mk #[f.toFloat8E4M3Bits]
+    -- float64 -> float8_e4m3
+    | .float64, .float8_e4m3 => do
+      let f <- Float.ofLEByteArray data
```
seanmcl, line 787: `float64 → fp8` casts via `.toFloat32` first (`f.toFloat32.toFloat8E4M3Bits`), which double-rounds and can disagree with ml_dtypes at the overflow boundary. Concretely, `float64` `464.00000000000006` (bits `0x407D000000000001`): ml_dtypes casts directly and yields **NaN** (byte 127), but here it rounds to `float32(464.0)` then to byte 126 (**448.0**, finite) — a finite-vs-NaN flip. (Also plain precision cases, e.g. `400.000000005` → 384 here vs 416 in ml_dtypes.) The existing `float64 → float16` path double-rounds the same way, so this is a pre-existing convention rather than new — but for fp8 the consequence crosses the NaN boundary. Flagging in case matching ml_dtypes exactly matters; otherwise it's a known tradeoff worth a one-line comment.

TensorLib/Float.lean
SmoothThunk: Updated comment to match the code.

TensorLib/Float.lean
SmoothThunk: Updated decoder's NaN comment about sign not being preserved.

TensorLib/Float.lean
SmoothThunk: Removed dead if check + added comment mentioning why this is unreachable.

TensorLib/Npy.lean
SmoothThunk: Added a comment for this

TensorLib/Npy.lean
SmoothThunk: Added these guard tests after bf16 guards.

TensorLib/Dtype.lean
SmoothThunk: Added decodeFloat8E4M3n / encodeFloat8E4M3 helpers with the size check in the decoder. Call sites have been updated to use these helpers

TensorLib/Dtype.lean
SmoothThunk: Added a comment for double rounding

SmoothThunk:
Also added .float8_e4m3 (used helper defined in dtype.lean) arm to toFloat32Tree and toFloat64Tree in Tensor. Its not a private function since I need to access it in tensor.lean.

SmoothThunk:
### `join` returns `none` for fp8_e4m3 + fp16 / bfloat16

`TensorLib/Dtype.lean:145` — the `.float8_e4m3, _ => none` catch-all swallows the fp16 and bfloat16 cases:

```lean
| .float8_e4m3, .float32 => float32
| .float8_e4m3, .float64 => float64
| .float8_e4m3, .bool
| .float8_e4m3, .int8
| .float8_e4m3, .uint8 => float8_e4m3
| .float8_e4m3, _ => none    -- fp16 and bfloat16 fall in here
```

But ml_dtypes / numpy promote these:

- `np.result_type(ml_dtypes.float8_e4m3fn, np.float16)` → `float16`
- `np.result_type(ml_dtypes.float8_e4m3fn, ml_dtypes.bfloat16)` → `bfloat16`

And the `lossless` table at `Dtype.lean:255` already asserts fp8 → fp16 and fp8 → bf16 are lossless, so declaring them unjoinable is self-inconsistent.

Suggested fix — add before the catch-all:

```lean
| .float8_e4m3, .float16 => float16
| .float8_e4m3, .bfloat16 => bfloat16
```

SmoothThunk:
ml_dtypes and numpy raise dtype promotion error with fp8_e4m3 , float16/ bfloat16 which matches current lean implementation so no changes.

SmoothThunk:
Reviewed the fp8_e4m3 diff against `ml_dtypes` at boundaries (0, ±0, subnormal thresholds, 240/256 rounding, 448/464 saturation, 465+ NaN, ±inf, integer overflow, join promotion rules). Could not confirm any correctness bugs — the encoder/decoder round-trip property and arithmetic outputs match `ml_dtypes` byte-for-byte.

Three lower-severity items:

- **TensorLib/Test.lean:288** — \`decode\` uses raw \`arr.data.data[offset]!\` instead of \`arr.data.extract offset (offset+1)\` like the fp16/bf16 tests (lines 64, 158). Works today because \`Tensor.ofNpy\` sets \`startIndex := 0\`, but diverges from the surrounding pattern and would silently break if that assumption ever changes.
- **TensorLib/Float.lean:289** — comment reads \`→ (1+7/8)*2^7 = 240? No...\` where 240 IS in fact correct (it just isn't the overall e4m3 max, since exp=15 with mant 0..6 gives values up to 448). The \`? No...\` phrasing misleads future readers.
- **TensorLib/Float.lean:264 / Dtype.lean:662** — small typos (\`inpit\` → \`input\`, \`atleast\` → \`at least\`).

SmoothThunk:
## Fixes:
* Test decode uses arr.data.extract through decodeFloat8E4M3 like in fp16 / bf16. This is safer because extract slices by position while the older data[offset]! assumed the tensor starts at 0 which would break for tensor slices
* Fixed typos in comments.

TensorLib/Dtype.lean
```diff
@@ -314,6 +333,16 @@ private def canCastFromInt (dtype : Dtype) (n : Int) : Bool :=
...
 def sizedStrides (dtype : Dtype) (s : Shape) : Strides := List.map (fun x => x * dtype.itemsize) s.unitStrides
 
+-- Decode 1-byte fp8_e4m3 to fp32
+-- Prevent having to repeat size check at call sites
+def decodeFloat8E4M3 (arr : ByteArray) : Err Float32 :=
+  if arr.size != 1 then .error "decoder: expected 1 byte for float8_e4m3"
+  else .ok (arr.data[0]!.toFloat32FromFloat8E4M3)
+
+-- Encode Float32 to 1-byte fp8_e4m3
+def encodeFloat8E4M3 (f : Float32) : ByteArray :=
```
seanmcl: Nit (non-blocking): `encodeFloat8E4M3` has no callers outside this module (all 11 uses are internal; `Tensor.lean` only calls `decodeFloat8E4M3`), and its direct sibling `encodeFloat16OrBFloat16` (line 476) is `private`. Consider marking this `private` to match and keep the encoder off Dtype's public surface. `decodeFloat8E4M3` correctly stays public since `Tensor.lean` calls it (mirroring the public `byteArrayToFloat16`/`byteArrayToBFloat16`).

seanmcl (commented):
Re-reviewed at the new head (`fa72f9e`) — thanks for the quick turnaround. I confirmed all six items from the last pass are correctly addressed, and the changes build clean with all `#guard`s passing:

- ✅ **`Tensor.lean`** — `toFloat32Tree`/`toFloat64Tree` now have explicit `.float8_e4m3` arms. I traced the leaf path (`toByteArrayTree` → `getPosition` slices `itemsize` = 1 byte for fp8), so `decodeFloat8E4M3`'s size guard passes and fp8 tensors decode correctly. The `Float32.toFloat` widening in `toFloat64Tree` is exact.
- ✅ **`decodeFloat8E4M3`/`encodeFloat8E4M3` helpers** — all former inline sites converted, none missed. I verified `encodeFloat8E4M3 f = ByteArray.mk #[f.toFloat8E4M3Bits]` is byte-identical to the previous forms (the `ToLEByteArray UInt8` instance is `ByteArray.mk #[n]`), so this is a pure extraction with no behavior change. Re-ran the numeric fuzz against `ml_dtypes` after the dead-code removal: still 0 mismatches over 3M+ values.
- ✅ **`abs`/`isZero`** — the removed inline `size != 1` guards are fully replaced by the helper's guard; the `.isOk == false` `#guard`s still hold.
- ✅ **`Float.lean`** — the dead `finalExp == 15 && finalMant == 7` branch is gone (I re-confirmed it's genuinely unreachable), and both comments (line 227 NaN semantics, line 251 NaN sign) are now accurate.
- ✅ **`Npy.lean`** — the `<V1`/`|V1` collision `#guard`s are present and pass, matching the bf16 `V2` precedent.

Nothing blocking from me. One tiny consistency nit inline; the pre-existing float64→fp8 double-rounding is now documented, which is fine. LGTM.

seanmcl (approved):

TensorLib/Dtype.lean
SmoothThunk: Fixed

PR #96, opened 2026-07-17
fp8_e5m2 dtype support

SmoothThunk:
Adds support for fp8_e5m2 (1 sign + 5 exponent bits + 2 mantissa)
* Bias = 15; contains inf and NaN
* Decoder handles 0, subnorms, +-inf, and NaN
* Encoder uses round to nearest even; overflow -> +- inf
* Helper functions contain a size guard and are used in all the arithmetic pattern matches, isZero, shift, etc
* Join was split into 2 functions join and joinOrdered to eliminate recursion since leans compiler timed out on a larger match.
* Npy parsing uses "<f2"
* Added tests for fp8_e5m2


SmoothThunk:
## Code review findings

Ranked most-severe first.

### 1. `TensorLib/Tensor.lean:603` — `toFloat32Tree`/`toFloat64Tree` miss `float8_e5m2`
The match in `toFloat32Tree` (line 597-603) has an explicit branch for `.float8_e4m3` but falls through to `Float32.ofLEByteArray` for e5m2. That helper calls `arr.toUInt32LE`, which requires 4 bytes and throws `"Expected size 4 byte array"` on a 1-byte e5m2 element. `toFloat64Tree` (line 607-614) has the same hole — its `_ =>` arm calls `Float.ofLEByteArray` (needs 8 bytes). Printing/decoding an e5m2 tensor is broken.

### 2. `TensorLib/Dtype.lean:964` — `liftFloatUnop` has no e5m2 case
`liftFloatUnop` matches on `.float8_e4m3`, `.float16 | .bfloat16`, `.float32`, `.float64` — no e5m2 arm. Any call to `Dtype.sin/cos/tan/exp/log/sqrt/arctan/tanh` on e5m2 hits the `_ => throw "operation requires a float type"` at line 981, even though `isFloat` (line 114) now claims e5m2 is a float.

### 3. `TensorLib/Npy.lean:98` — `"f1"` decoded without a byte-order guard
For bfloat16 the code explicitly requires `<V2` (rejecting `|V2`) to avoid collision with generic void data, and `<V1` is required for e4m3. The new `"f1" => float8_e5m2` mapping in `dtypeNameFromNpyString` is reached for any byte-order prefix (`<f1`, `>f1`, `|f1`, `=f1`). NumPy has no native `f1`, so an external producer writing `|f1` for a custom 1-byte float will be silently reinterpreted as e5m2. Inconsistent with the guard rule used for e4m3.

### 4. `TensorLib/Dtype.lean:204` — `join` relies on both argument orders being listed in `joinOrdered`
`join` only swaps when `x.itemsize > y.itemsize`. For equal-itemsize pairs (all the 1-byte types — bool, int8, uint8, float8_e4m3, float8_e5m2), no swap happens, so both `(x,y)` and `(y,x)` reach `joinOrdered` in their original order. Any missing direction silently returns `none` instead of promoting. Today the coverage looks right, but this is a latent commutativity bug — consider normalizing argument order (e.g. sorting by dtype tag) before calling `joinOrdered`, or adding bidirectional promotion tests.

### 5. `TensorLib/Float.lean:371` — subnormal e5m2 decode uses fp32 multiplication (simplification)
`UInt8.toFloat32FromFloat8E5M2` handles subnormals by `Float32.ofNat mant * 2^-16` and then bit-ORing the sign. Correct today, but the normal path two branches below (line 387-388) constructs the fp32 bits directly. Uniform bit-construction would be simpler and avoids any dependence on Float32 multiplication semantics. Same pattern in the encoder subnormal path at line 464-474.

### 6. `TensorLib/Dtype.lean:127` — truncated comment on `joinOrdered`
The rationale comment reads `"...adding more fp8 types made the match too large for — Lean needs to prove termination..."`. The em-dash breaks the sentence — appears to have dropped a word. Load-bearing context (explains why join was refactored) that will confuse future readers if left as-is.

SmoothThunk:
## Fixes:
* Fp8_e5m2 arms already handled.
* Added fp8_e5m2 case to liftFloatUnop
* Only <f1 (littleEndian) is now recognized as fp8_e5m2, matching the guard pattern from previous dtypes. Added |f1 rejection guard.
* Added a PBT for join commutativity that catches any missing pairs at build time.
* I use multiplication because Lean's fp32 is opaque - multiplying mant * 2^(-16) lets fp32 handle the encoding correctly without manually constructing subnormal bit patterns. Same approach as e4m3/fp16 decoders, verified by PBT.
* fixed comment

SmoothThunk:
## Adversarial code review — findings

Ran a follow-up adversarial pass focused on inputs the earlier reviews didn't probe (rounding boundaries, NaN/inf casts, silent truncation, joinOrdered exhaustively vs `np.result_type`). Most angles came back clean; one real bug surfaced, with two failure modes sharing the same root cause.

### 1. e5m2 ±inf → signed int returns wrong value (correctness)

**Where:** `TensorLib/Dtype.lean:853-858` (e5m2 → signed int cast) via `TensorLib/Float.lean:88` (`Float32.toInt`).

**Failure:** `castOverflow .float8_e5m2 (bytes [252]) .int8` returns `[1]`. Numpy returns `0` for `-inf → int8`.

Path: `decodeFloat8E5M2 0xFC` produces fp32 `−∞`; `Float32.toInt` at Float.lean:88 does `f.toUInt64.toNat`, which for `±inf` returns `2^64 − 1` regardless of sign; the negative branch then returns `−(2^64 − 1)`; `byteArrayOfIntOverflow .int8` truncates mod 256 → `1`.

### 2. e5m2 ±inf → int32/int64 also wrong (same root cause)

`castOverflow .float8_e5m2 (bytes [124]) .int32` returns `0xFFFFFFFF` (=−1). Numpy saturates to `INT32_MAX = 2147483647`. Same for int64. NaN → int returns 0 (correct by accident — `NaN.toUInt64 = 0`).

Both cases are the same underlying `Float32.toInt`/`Float32.toNat` bug at `TensorLib/Float.lean:86-92`. It is pre-existing (fp32/fp16/bf16 → int on inf/NaN already diverged from numpy), but the e5m2 diff makes it newly reachable via a 1-byte dtype since e4m3fn has no infinities and never surfaces it.

**Suggested fix:** make `Float32.toInt`/`.toNat` saturate to `INT64_MIN`/`MAX` for `±inf` and return 0 for NaN, matching numpy. That also cleans up the pre-existing divergence for the other float dtypes' inf/NaN → int paths.

**Test gap enabling this:** `testFloat8E5M2EdgeCases` covers inf → e5m2 encoding but never inf-e5m2 → int decoding, NaN decode, or e5m2 → e4m3. Adding inf/NaN → int cases would have caught this.

### Angles that came back clean

- Silent `UInt8` truncation in the encoder — traced all `.toUInt8` casts: `finalExp ≤ 30`, `finalMant ≤ 3`, subnormal `rounded < 4`. No path can overflow.
- RTNE halfway cases at 61440 (inf boundary), 1.5·2^-16 (subnormal↔normal), 2^-17 (0↔subnormal), and normal-range halfway values — all match ml_dtypes.
- Double-rounding fp64 → fp32 → e5m2 vs direct fp64 → e5m2 in the overflow region — no divergence.
- Subnormal encoder algebra (`totalShift = 7 − realExp`) — verified by hand for realExp ∈ {−15..−31}.
- `joinOrdered` — exhaustively checked all `e5m2 × T` and `T × e5m2` against `np.result_type`, including `e5m2 + e4m3` and `e5m2 + bfloat16` correctly returning `none` (numpy raises `DTypePromotionError`).
- NPY `<f1`-only acceptance — matches numpy's own emitted string; `|f1` and `=f1` are unparseable in numpy, so rejecting them is correct.

SmoothThunk:
### Fixes: 
* Float32.toInt saturates to INT64_MAX for +inf and INT64_MIN for -inf, returns 0 for NaN - matching numpy's int64 saturation behavior. byteArrayOfIntOverflow then truncates into the target type (e.g. INT64_MAX mod 256 = -1 for int8, INT64_MAX for int64). Added #guard tests. This also fixes the pre-existing divergence for fp16/bf16/fp32 -> int on inf/NaN. Also added IO tests for inf/NaN -> int8 and e5m2 -> e4m3 casts.

SmoothThunk:
## Code review — fix commit follow-up

Reviewing 4f4ebf1 (Float32.toInt inf/NaN fix). The direction is right, but the fix is incomplete and the chosen sentinel values only match numpy at int8/int16 by coincidence.

### 1. Float32.toInt hard-codes -1 / 0 instead of saturating (correctness, high)
**`TensorLib/Float.lean:91`**

The fix returns `-1` for +inf and `0` for -inf regardless of the target int size. That happens to match numpy for int8/int16, but diverges for int32/int64:

- `castOverflow .float8_e5m2 [124] .int32` (e5m2 +inf → int32): Lean returns `-1`, numpy returns `INT32_MAX = 2147483647`.
- `castOverflow .float8_e5m2 [124] .int64`: Lean returns `-1`, numpy returns `INT64_MAX = 9223372036854775807`.
- `castOverflow .float8_e5m2 [252] .int32` (-inf → int32): Lean returns `0`, numpy returns `INT32_MIN`.
- Same divergence for -inf → int64.

The earlier suggestion was to return `INT64_MIN`/`INT64_MAX` and let `byteArrayOfIntOverflow`'s modular truncation produce numpy's saturation at each smaller size:

- `INT64_MAX.toInt8 = -1` (matches numpy int8)
- `INT64_MAX.toInt16 = -1` (matches numpy int16)
- `INT64_MAX.toInt32 = -1`... which still diverges from numpy's `INT32_MAX`.

So truncation alone doesn't recover numpy's per-size saturation either — the correct fix probably has to happen inside `byteArrayOfIntOverflow` (or a new saturating variant), keyed on the target dtype. Reference behavior:

```
>>> np.array([np.inf, -np.inf, np.nan], dtype=np.float32).astype(np.int32)
array([ 2147483647, -2147483648,           0], dtype=int32)
```

### 2. Float32.toNat was not updated (correctness, high)
**`TensorLib/Float.lean:86`**

`Float32.toNat` still does `f.toUInt64.toNat` unchanged — no inf/NaN handling. Any `float → uint*` cast on ±inf/NaN still hits the pre-existing bug: `Dtype.lean:765/785/815/851` all route unsigned float→int casts through `f.toNat`. Numpy's behavior:

- +inf → uint8/16/32/64 = UINT_MAX for that size (255 / 65535 / …)
- -inf → uint* = 0
- NaN → uint* = 0

### 3. Float.toInt / Float.toNat (fp64) untouched (correctness)
**`TensorLib/Float.lean:102-108`**

The fp64 helpers still have the pre-existing bug shape (`neg := f <= 0; -f; toUInt64.toNat`). Reachable via `float64 → int8/int16/int32/int64` casts (`Dtype.lean:772-774`) on ±inf/NaN inputs.

### 4. Tests only cover int8 (test-coverage)
**`TensorLib/Test.lean:526-549`**

The new tests exercise e5m2 (+inf / -inf / NaN) → int8 and e5m2 +inf → e4m3. Since -1 truncates the same way at int8 (0xFF = -1) regardless of the underlying saturation choice, this coverage doesn't distinguish the -1-sentinel behavior from the INT_MAX-saturation behavior. Adding equivalent tests for int32, int64, and any of the uint targets would surface #1 and #2.

### 5. Consider extracting Float32.isNaN / isPosInf / isNegInf (simplification)
**`TensorLib/Float.lean:89-91`**

Three separate special-case branches (`f != f`, bit-equality with `0xFF800000`, bit-equality with `0x7F800000`) are fragile — Lean's Float32 `==` on infinities is IEEE-defined (true for identical inf), but it's worth encoding that assumption as helpers plus a `#guard Float32.ofBits 0x7F800000 == Float32.ofBits 0x7F800000` sanity check so the invariant is executable rather than implicit.

seanmcl (approved):

SmoothThunk:
## Fixes:
* Fixed. Added intMin/intMax accessors and saturatingIntOfFloat32/saturatingIntOfFloat64 helpers that saturate to the target dtype's bounds for +-inf and return 0 for NaN. Rewired all 5 float -> int cast sites to use the helpers. Now matches numpy for all int sizes (+inf -> int8 = 127, +inf -> int32 = INT32_MAX, +inf -> int64 = INT64_MAX, etc). Updated tests accordingly.
* Fixed. Fp32.toNat now explicitly handles NaN -> 0, negatives/-inf → 0, and +inf → UINT64_MAX. Truncation in byteArrayOfNatOverflow gives the correct UINT_MAX per target size (e.g. 255 for uint8, 65535 for uint16). Added #guard tests.
* Fixed; applied the same inf / NaN handling to Float.toNat and Float.toInt (fp64 versions). Added #guard tests for both.
* Added tests for uint8 and int32 targets. Uint cases (+inf -> 255, -inf -> 0) match numpy. Int32 case documents the divergence (+inf gives -1, numpy gives INT32_MAX)
* Extracted isPosInf/isNegInf helpers for both fp32 and fp64. Updated toNat/toInt for both fp32 and fp64 to use
  f.isNaN/f.isPosInf/f.isNegInf. Added #guard checks for IEEE 754 inf equality.

SmoothThunk:
Update after implementing the fix — a self-review turned up 6 issues worth flagging. Top one is a real gap:

**1. `saturatingIntOfFloat*` doesn't actually saturate finite overflow.** The helper only saturates ±inf/NaN; a finite `1e10 → int32` still wraps via `.toInt32` (yielding 1410065408 instead of numpy's INT32_MAX = 2147483647). The name and the `intMin`/`intMax` comment ("we can saturate to INT32_MAX ...") overpromise vs. what the code delivers. Fix is a range clamp before falling through to `f.toInt`.

**2. `Float.toInt` / `Float.toNat` missed the refactor.** They still use `f != f` for NaN and raw `f == Float.ofBits 0xFFF0...` for -inf, while the Float32 siblings were switched to the new `isNaN` / `isPosInf` / `isNegInf` helpers introduced in the same diff. The comment at `Float.lean:128` also references a "TODO comment above" that doesn't exist.

**3. `intMin` / `intMax` wildcard swallows misuse.** `| _ => 0` returns 0 for uint/float/bool. If a future refactor folds uint saturation into `saturatingIntOfFloat32`, `-inf → uint8` silently yields 0 with no signal.

**4. `Float32.toNat` / `Float.toNat` handle +inf only implicitly** via `f.toUInt64.toNat`, relying on unspecified Lean runtime behavior. `toInt` handles +inf explicitly; `toNat` should too for symmetry.

**5. Stale comment at `Test.lean:526`** — "INT64_MAX truncated to int8" describes the old path; new value 127 comes directly from `intMax .int8`.

**6. Tautological comment at `Test.lean:568`** — "we give 2147483647, numpy gives INT32_MAX" implies divergence, but those are the same value.

Will address 1–4 before merging; 5 and 6 are comment-only fixes.

SmoothThunk:
## Fixes:
* Added saturation clamp in byteArrayOfIntOverflow for int32/int64 - finite overflow now saturates to INT32_MAX/MIN matching numpy. int8/int16 keep truncation (also matches numpy's wrapping behavior for those sizes).
* Float.toNat and Float.toInt now use isNaN/isPosInf/isNegInf helpers, matching the Float32 versions. Also removed old comment.
* Keeping | _ => 0; listing all non-int types explicitly adds clutter without changing behavior. The wildcard keeps it concise.
* Both Float32.toNat and Float.toNat now handle +inf explicitly instead of depending on toUInt64's unspecified behavior.
* Refactored comment

SmoothThunk:
Self-review of `dea54d1` (finite-overflow clamp in `byteArrayOfIntOverflow`) — 5 findings:

**1. `byteArrayOfIntOverflow` is shared with integer arithmetic — this regresses numpy parity for int ops.** `sub` (line 631), `mul` (line 663), `div` (line 695), and `abs` (line 729) all call it for signed-int dtypes. Before this commit those paths wrapped on overflow (matches numpy int arithmetic). After this commit, `.int32` and `.int64` **saturate**. Example: `int32(INT32_MAX) * int32(2)` previously produced `-2` (wrap, `np.int32(2**31-1) * np.int32(2) == -2`); now produces `2147483647`. This regresses parity in the opposite direction from what the fix targeted. Biggest issue.

**2. `int8` / `int16` still wrap; only `int32` / `int64` clamp.** `float32(1e10) → int8` still goes through `(10000000000 : Int).toInt8` (mod 256 noise byte) instead of numpy's 127. The finite-overflow parity fix is only half-applied.

**3. `float → uint` casts still wrap entirely.** All uint cast paths call `byteArrayOfNatOverflow` (unchanged). `float32(1e20) → uint32` wraps instead of saturating to `UINT32_MAX`. `+inf → uint*` happens to give `UINT_MAX` per width via mod on `Float32.toNat(+inf) = 0xFFF...F`, but finite overflow diverges.

**4. `saturatingIntOfFloat32/64` helpers are now partially redundant.** With finite saturation moved into `byteArrayOfIntOverflow`, the helper's `else f.toInt` branch is dominated by the clamp downstream. Its real responsibility is now only inf/NaN handling; the name overpromises.

**5. `intMin` / `intMax` wildcard `| _ => 0` still open** from the previous review round.

**Suggested direction:** the shared helper is fighting two callers with opposite semantics (arithmetic wants wrap, casts want saturate). Options:
- Split into `byteArrayOfIntWrap` (arithmetic) and `byteArrayOfIntSaturate` (casts), or
- Push saturation back to `saturatingIntOfFloat*` where the caller identity is clear, and revert `byteArrayOfIntOverflow` to pure wrap.

The second is smaller and matches the helper's name (`Overflow` reads as "wrap on overflow" today).

SmoothThunk:
## Fixes:
* Fixed; Moved saturation from byteArrayOfIntOverflow (shared with integer arithmetic) into saturatingIntOfFloat32/saturatingIntOfFloat64 only. Integer arithmetic now wraps again (matching numpy), while float -->int casts still saturate. Added min (max f.toInt dtype.intMin) dtype.intMax for finite overflow in the helpers.
*  Finite overflow has been updated to only saturate for int32/int64. int8/int16 wrap via truncation. Numpy's behavior is inconsistent across sizes due to C cast semantics.
* Added saturating helpers for unsigned casts  so now it saturates uint32/uint64, wraps uint8/uint16 (matches numpy). Rewired all 5 float -> unsigned cast sites. byteArrayOfNatOverflow stays unchanged so integer arithmetic still wraps.
* The saturation is in the helpers. The helpers handle inf/NaN and finite overflow clamping for int32/int64. The name is accurate. byteArrayOfIntOverflow is back to pure truncation so integer arithmetic wraps correctly.
* Keeping | _ => 0 keeps it concise and the saturating helpers only pass signed/unsigned int dtypes. Adding explicit cases for all fp types adds clutter without changing behavior.
* Suggested direction: byteArrayOfIntOverflow is pure wrap which is used by arthmetic. Saturation is a different function (used only by cast paths).

SmoothThunk:
Self-review of `dd4794d` (saturation moved into `saturatingIntOfFloat*` / new `saturatingNatOfFloat*`) — 7 findings. Architectural direction is much better than the previous commit (saturation is out of the shared arithmetic helper), but there's a real bug on the fp64 lane.

**1. `saturatingIntOfFloat64` has the wrong match arms — fp64→int32 still wraps.** Line 474 clamps on `| .int64 | .uint64 =>`, but this function returns `Int` and is only called from float64→**signed** int paths (line 844). `.uint64` is dead code; `.int32` is missing. Result:

- `castOverflow .float64 (toLEByteArray (1.0e10 : Float)) .int32` falls through to `| _ => f.toInt`, then `byteArrayOfIntOverflow` (which no longer clamps) yields `1410065408`. Numpy gives INT32_MAX.
- `saturatingIntOfFloat32` at line 464 has it right (`| .int32 | .int64 =>`), so fp32→int32 works. Silent divergence between fp source widths for the same target.

Should be `| .int32 | .int64 =>`.

**2. uint32/uint64 saturation of large-finite depends on unspecified `Float.toUInt64` behavior.** Lines 440/452: for `f > UINT64_MAX` (e.g. `float32(1e30)`), `f.toNat` bottoms out on Lean's `Float32.toUInt64.toNat` for out-of-range inputs — undefined, platform-dependent (INDEFINITE on x86-family). `min f.toNat dtype.intMax.toNat` then clamps whatever came back. Same class of bug the earlier `+inf → uint*` fix addressed with an explicit `isPosInf`; here the "finite huge" case has no guard.

**3. Negative finite → uint8/uint16 contradicts its own docstring.** Line 430 says "wrap for uint8/uint16" and "Negative finite: wrap for uint8/uint16". But the code falls through to `| _ => f.toNat`, and `Float32.toNat` returns 0 for `f <= 0`. So `float32(-1.5) → uint8 = 0`, not the wrap value 255/254. Numpy: 255. Doc and code disagree; both diverge from numpy.

**4. `.uint64` arm in `saturatingIntOfFloat64` is dead code.** Follows from #1 — this function only handles signed targets, so `.uint64` never matches. Uint64 casts go through `saturatingNatOfFloat64`.

**5. Overconfident numpy-parity claim.** Line 430 says uint8/uint16 wrap "also matches numpy's C cast behavior". Numpy's float→uint for out-of-range is technically undefined (RuntimeWarning), platform-dependent. Combined with #3, the "matches numpy" claim is unjustified for the negative-uint8 case.

**6. Asymmetric arm lists between siblings.** `saturatingIntOfFloat32` lists `.int32 | .int64`; `saturatingIntOfFloat64` lists `.int64 | .uint64`. Maintenance trap even setting aside #1 — bug fix here should probably keep the two mirroring each other.

**7. Typo at line 456.** `satuate` → `saturate`.

SmoothThunk:
## Fixes:
* Changed saturatingIntOfFloat64 match arms to | .int32 | .int64.
*  Added guard f.toFloat >= 2^64 before calling toNat so values exceeding UInt64 range saturate to dtype.intMax.toNat instead of hitting undefined toUInt64 behavior. Applied to both saturatingNatOfFloat32 and saturatingNatOfFloat64.
* Negative finite -> uint8/uint16 now wraps via f.toInt % 256 / f.toInt % 65536 (matching numpy). Updated bf16 test from == 0 to == 255 and related comment.
* fixed; saturatingIntOfFloat64 now has | .int32 | .int64 => with no .uint64.
* Updated comment
* Fixed in prev comment for this PR.
* fixed spelling

SmoothThunk:
Self-review of `3060e30` — 6 findings with suggested fixes. Previous bug (`saturatingIntOfFloat64` missing `.int32`) is fixed; architecture is settling. Remaining issues below.

**1. `>= 2^64` guard is not tight for the uint64 target.** Lean's `Float.toUInt64` is `@[extern]` C, which on x86 lowers to CVTTSS2SI (signed 64-bit) and is UB for values in `[2^63, 2^64)` — typically returns `0x8000000000000000`. For `.uint32` the `min(_, UINT32_MAX)` clamp masks this; for **`.uint64`**, `min(2^63, UINT64_MAX) = 2^63`, so `float32(1.8e19) → uint64` silently returns `2^63` instead of saturating.

**Suggested fix** — key the threshold on target width:

```lean
| .uint32 | .uint64 =>
  if f <= 0 then 0
  else if f.toFloat >= 9223372036854775808.0 then dtype.intMax.toNat  -- 2^63, boundary of toUInt64 UB
  else min f.toNat dtype.intMax.toNat
```

For `.uint32`, values in `[2^31, 2^63)` are still fine because `f.toUInt64` returns the correct value there (fits in signed 64-bit range); `min` then clamps to UINT32_MAX. For `.uint64`, anything `>= 2^63` now short-circuits to UINT64_MAX. Apply the same change to `saturatingNatOfFloat64` at line 463.

**2. Dead code: `if n < 0` branch after `Int.%`.** Lean's `Int.%` is `Int.emod` (Euclidean modulo — see `Init/Data/Int/DivMod/Basic.lean:116`), which returns non-negative results for a positive divisor. The `if n < 0 then (n + 256).toNat else n.toNat` branches at lines 447, 450, 466, 469 can never take the true branch.

**Suggested fix:**

```lean
| .uint8  => (f.toInt % 256).toNat
| .uint16 => (f.toInt % 65536).toNat
```

(Same for fp64 helper.)

**3. uint8/uint16 wrap path exposes `toUInt64` UB for large-magnitude finite.** `f.toInt` for `|f| >= 2^63` calls `f.toUInt64` under the hood, which is UB. Then `% 256` produces platform-dependent bytes. Numpy is also undefined here, so this is between-platforms nondeterminism, not against-numpy divergence — but the "we match observed x86 behavior" note in the docstring won't survive a compiler-lowering change.

**Suggested fix** — same threshold trick: bail out to a fixed value before hitting `f.toInt`.

```lean
| .uint8 =>
  if f.toFloat.abs >= 9223372036854775808.0 then 0  -- or dtype.intMax.toNat; document choice
  else (f.toInt % 256).toNat
```

Numpy's undefined behavior here means there's no "correct" answer — pick a deterministic one and document it. Zero is defensible (matches float→bool convention for out-of-range).

**4. No tests for the new finite-saturation paths.** The only new/changed test asserts `-1.5 → uint8 = 255`. Missing coverage: `float32(1e10) → uint32`, `float32(1e10) → uint64`, `float32(-1e10) → uint32` (clamp to 0), and the boundary case `float32(1.8e19) → uint64` from finding #1.

**Suggested fix** — add to `testFloat8E5M2EdgeCases` or a new `testFloat32SaturationEdgeCases`:

```lean
-- finite overflow: uint32 saturates
let bigF32 := toLEByteArray (1.0e10 : Float32)
let castBig <- IO.ofExcept (Dtype.castOverflow .float32 bigF32 .uint32)
let pass := castBig.toNat == 4294967295
checks := pass :: checks

-- boundary case exposing toUInt64 UB
let hugeF32 := toLEByteArray (1.8e19 : Float32)
let castHuge <- IO.ofExcept (Dtype.castOverflow .float32 hugeF32 .uint64)
let pass := castHuge.toNat == 18446744073709551615
checks := pass :: checks

-- negative large: clamp to 0 for uint32/64
let negF32 := toLEByteArray (-1.0e10 : Float32)
let castNeg <- IO.ofExcept (Dtype.castOverflow .float32 negF32 .uint32)
let pass := castNeg.toNat == 0
checks := pass :: checks
```

**5. Asymmetry between fp32 and fp64 helpers.** Line 442 uses `f.toFloat >= 2^64` (fp32 promoted), line 463 uses `f >= 2^64` directly. Not a bug; copy-paste hazard.

**Suggested fix** — same threshold constant, factored out:

```lean
private def uintOverflowThreshold : Float := 9223372036854775808.0  -- 2^63, boundary of toUInt64 UB
```

Then both helpers reference the same constant. Or inline the comment `-- 2^63: boundary of Float.toUInt64 UB` at both sites.

**6. Docstring wording.** Lines 429–432: correct but slightly awkward. Suggest:

```lean
-- Saturating fp → Nat, keyed on target unsigned dtype.
-- +inf -> uintMax, -inf -> 0, NaN -> 0.
-- Positive finite: saturate to uintMax for uint32/uint64; wrap mod 2^N for uint8/uint16.
-- Negative finite: clamp to 0 for uint32/uint64; wrap mod 2^N for uint8/uint16.
-- Note: numpy's float -> uint out-of-range is platform-dependent (RuntimeWarning).
-- We match observed x86-64 behavior.
```

Fixes 1 and 2 are the ones worth doing before merging; 3–6 are polish.

SmoothThunk:
## Fixes:
* Fixed- Tightened the overflow guard to >= 2^63 in both saturatingNatOfFloat32 and saturatingNatOfFloat64.
* Removed dead if n < 0 branches — Lean's Int.% is Euclidean (always non-negative for positive divisor). Simplified to (f.toInt % 256).toNat. Added #guard tests to verify the cast for -fp -> uint8.
* . Added >= 2^63 / <= -2^63 guards before calling f.toInt in the uint8/uint16 paths. Large positive → UINT_MAX for that size, large negative -> 0. Matches x86 behavior. Added #guard tests for both.
* Added tests for finite overflow.
* Extracted constant (2^63) to reduce any typos
* Updated comment.

SmoothThunk:
## Code Review — `6d4cfce`

Small diff (49 Dtype + 14 Test) that addresses **all four correctness findings** from the previous review (threshold tightened to 2^63, dead `if n < 0` code removed, uint8/uint16 large-magnitude UB avoided via early sentinel, new saturation tests added). Traced by hand — all test assertions match the code paths. Remaining 5 findings are mostly documentation and coverage polish.

**1. Docstring vs code mismatch on large-magnitude negatives** (`Dtype.lean:433`). Docstring says `Negative finite: ... wrap mod 2^N for uint8/uint16`, but the new uint8/uint16 code (lines 449, 453) **clamps** large-magnitude negatives (`f <= -2^63`) to 0 — only wrapping for `-2^63 < f < 0`. The undocumented threshold is 2^63; nothing in the docstring reveals the split.

Suggested rewording:

```lean
-- Negative finite in (-2^63, 0): wrap mod 2^N for uint8/uint16, clamp to 0 for uint32/uint64.
-- Large magnitude (|f| >= 2^63): fixed sentinel (uintMax for positive, 0 for negative)
-- to avoid Float.toUInt64 UB. Numpy is also undefined here.
```

**2. `"We match observed x86-64 behavior"` overclaims parity** (`Dtype.lean:435`). For `|f| >= 2^63` the code returns fixed sentinels (255/65535 for positive, 0 for negative). Numpy on those inputs is undefined and typically involves platform-defined wraps, not clean 0/uintMax. The comment implies numpy parity we don't deliver in that range.

Suggested rewording:

```lean
-- Note: numpy's float -> uint out-of-range is technically undefined (RuntimeWarning).
-- For |f| < 2^63 we match numpy's typical x86-64 wrap behavior;
-- for |f| >= 2^63 we return deterministic sentinels (uintMax / 0) since Float.toUInt64 is UB there.
```

**3. No fp64 tests for the new `saturatingNatOfFloat64` paths** (`Test.lean:583`). The new tests at `Test.lean:569-581` and `Dtype.lean:1443-1448` cover only `float32` sources. `saturatingNatOfFloat64` at `Dtype.lean:458-476` has the same restructure but no test. Given fp32/fp64 drift has been the recurring bug pattern in this refactor (e.g., the earlier `saturatingIntOfFloat64` missing `.int32` case), this is worth closing.

Suggested additions:

```lean
-- fp64 mirror of the fp32 saturation tests
#guard Dtype.castOverflow .float64 (toLEByteArray (-1.0 : Float)) .uint8 == .ok (ByteArray.mk #[255])
#guard Dtype.castOverflow .float64 (toLEByteArray Float.maxValue) .uint8 == .ok (ByteArray.mk #[255])
#guard Dtype.castOverflow .float64 (toLEByteArray Float.minValue) .uint8 == .ok (ByteArray.mk #[0])

-- Finite overflow: float64(1e10) -> uint32 saturates
let bigF64 := toLEByteArray (Float.ofNat 10000000000)
let castBig <- IO.ofExcept (Dtype.castOverflow .float64 bigF64 .uint32)
let pass := castBig.toNat == 4294967295
```

**4. fp64 helper has a two-word docstring** (`Dtype.lean:457`). `-- Same for fp64` on a 20-line helper with the same restructured logic as fp32. A reader editing `saturatingNatOfFloat64` in isolation would miss the shared invariants (threshold value, sentinel choices).

Suggested:

```lean
-- Same contract as saturatingNatOfFloat32; see docstring there for details.
```

**5. uint8/uint16 arms are near-duplicates across fp32/fp64** (`Dtype.lean:447`). Four near-identical bodies differing only in modulus (256 vs 65536) and max value (255 vs 65535). Not a correctness bug, but the previous review round had exactly this duplication produce a real bug (`saturatingIntOfFloat64` missing `.int32`). A local helper would reduce future drift risk:

```lean
private def wrapModPow2 (bits : Nat) (f : Float32) : Nat :=
  let mask := (1 <<< bits) - 1
  let uintMax := mask
  if f.toFloat >= uintOverflowThreshold then uintMax
  else if f.toFloat <= -uintOverflowThreshold then 0
  else (f.toInt % (mask + 1)).toNat
```

Then `.uint8 => wrapModPow2 8 f` and `.uint16 => wrapModPow2 16 f`, with a fp64 twin. Optional cleanup; the current form is correct as-is.

---

Correctness-wise, the code is in good shape. The 2^63 threshold correctly bounds the UB range for both `Float.toUInt64` (positive side) and `Float.toInt` (which calls `toUInt64` internally on negative magnitudes). Merge-ready after the docstring fixes in 1 and 2; 3-5 are optional polish.

SmoothThunk:
## Fixes:
* Updated comment to document that values with |f| >= 2^63 return fixed values (uintMax or 0) instead of computing, to avoid undefined behavior.
* Clarified numpy parity only holds for |f| < 2^63; for |f| >= 2^63 we use deterministic sentinels where both numpy and Lean are undefined.
* Added #guard tests for fp64 -> uint8 (negative wrap, max, min) and IO test for fp64(1e10) -> uint32 saturation. Verified against numpy.
* Updated comment in saturation for fp64.
* Extracted wrapUintMod / wrapUintModF64 helpers to reduce code duplication.

SmoothThunk:
## Code Review — `bc340ff`

Clean refactor commit: extracts `wrapUintMod` / `wrapUintModF64` helpers to remove duplication, rewords docstring per prior finding #1, and adds fp64 saturation tests per prior finding #3. All previous review action items are addressed.

**Only 2 minor findings, both PLAUSIBLE — no confirmed correctness bugs.**

**1. Naming asymmetry** (`Dtype.lean:433`, cosmetic). The two extracted helpers are `wrapUintMod` (fp32) and `wrapUintModF64` (fp64). The rest of the file uses the full `Float32`/`Float64` spelling (`saturatingIntOfFloat32`/`saturatingIntOfFloat64`, `saturatingNatOfFloat32`/`saturatingNatOfFloat64`). Mixing `F64` and `Float64` conventions within the same subsystem invites future grep-based rename misses.

Suggested rename: `wrapUintModOfFloat32` / `wrapUintModOfFloat64` to match the neighbors.

**2. `wrapUintMod` accepts any `bits : Nat` with no validation** (`Dtype.lean:433`, latent risk). Callers today only pass 8 and 16, which is safe. If someone adds `.uint32 => wrapUintMod 32 f` naively (forgetting that uint32 saturates, not wraps), the helper would return `f.toInt % 2^32` — silently applying wrap-mod semantics where saturate-to-UINT32_MAX is intended. Private function, small risk today.

Suggested minor hardening: rename to `wrapUintMod8Or16` or add an assertion / doc note that this helper is only for widths ≤ 16.

---

**What I verified (didn't fire as findings):**

- **Precedence**: `1 <<< bits - 1` correctly parses as `(1 <<< bits) - 1` — `<<<` is `infixl:75`, `-` is `infixl:65`.
- **Guard endpoints**: `f.toFloat >= 2^63` / `f.toFloat <= -2^63` are inclusive on both sides, correctly bounding the `Float.toUInt64` UB range. On the negative side, `-2^63` is representable in fp32/fp64 but its negation `2^63` is UB for `toUInt64`, so `<=` (not `<`) is needed. Correct.
- **`f.toInt` in the wrap path** only sees `|f| < 2^63`, where `Float32.toInt`'s internal `f.toUInt64` call is well-defined.
- **Test values** (`Float.maxValue ≈ 1.8e308`, `Float.minValue ≈ -1.8e308`, `1e10`) all fall on the correct branches with the asserted outputs; traced by hand.
- **fp64 tests mirror fp32 tests** — closes the drift risk noted last round.

Correctness-wise this diff is in good shape. Both findings are polish; neither blocks merge.

SmoothThunk:
## Fixes:
* Updated helper function names to be consistent.
* Added comment clarifying the helper is only for widths <= 16. uint32/uint64 use the saturation path instead.

